PR #6, opened 2024-11-16
Add and run precommit hooks

Jacob-Chmura:
In this PR I have added a [precommit config](https://github.com/shenyangHuang/openDG/blob/dev/precommit/.pre-commit-config.yaml) and the associated [dev dependencies](https://github.com/shenyangHuang/openDG/blob/0ab82029e5bad35f910e99ecc55e69c8d9b91d4f/pyproject.toml#L17) to the project. Just so you know, I have installed the hooks locally, hence the changes to the source code.

We can discuss each hook in our next meeting.

---
Stacking #5

PR #7, opened 2024-11-16
Add workflows

Jacob-Chmura:
I've added some basic workflows for formatting and running tests on merges to master. The most up for debate is [static analysis](https://github.com/shenyangHuang/openDG/blob/dev/workflows/.github/workflows/mypy.yml). I've deliberately let it fail here for you to refer to.

---
Stacking #6


Jacob-Chmura:
The UV lock file is bricked. Closing and re-opening with a consistent environment state.

PR #8, opened 2024-11-19
Adding CTDG data structure

shenyangHuang:
Initial implementation of CTDG data structure. 
- see [`test_loading.py`](https://github.com/shenyangHuang/openDG/blob/data_dev/test/test_data/test_loading.py) for unit tests for the `EventStore` and `CTDG` objects 
- [x] `CTDG` object now supports loading from csv edgelists 
- [x] `CTDG` object now supports counting `num_edges`, `min_time`, `max_time`
- [x] `CTDG` object now contains `aggregate_graph` function and you can specify a time interval to aggregate over. 
- [x] `CTDG` object supports `to_events` function

Synthetic test with 10,000 edges, 10,000 nodes and 10,000 timestamps on a path graph takes 10 seconds on my laptop. 
Operations include: 1. load from csv, 2. `aggregate_graph` (twice) 3. `to_events`

To do:
- [ ] `CTDG` supports `to_snapshots` function


opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+            del self._store[key]
+
+    def __copy__(self):
+        return CTDG(self._store.copy())
+
+    def __deepcopy__(self, memo):
+        return CTDG(self._store.copy())
+
+    def __repr__(self) -> str:
+        return f"CTDG({self._store.keys()})"
```
Jacob-Chmura, line 201: I'm unsure how helpful this `__repr__` is for debugging, and since we don't override `__str__` it's what the user will see if they stringify the object. If we have thousands of edges, this will just bombard their terminals.

I don't really have a good solution, but we may look into how `numpy` does their `__repr__` with large arrays, for instance: https://github.com/numpy/numpy/blob/v2.1.0/numpy/_core/arrayprint.py#L1609-L1656

In short, among other things, they will compactify large arrays like this:
```python
>>> repr(np.array([i for i in range(1001)]))
'array([   0,    1,    2, ...,  998,  999, 1000])'
```

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+        t_list = list(edge_dict.keys())
+        if (not all(t_list[i] <= t_list[i+1] for i in range(len(t_list) - 1))):
+            edge_dict = dict(sorted(edge_dict.items()))
+
+        #! add logic to insert node events too
+        self.__dict__['_store']['events'] = self._to_eventStore(edge_dict)
+        ts_list = list(edge_dict.keys())
+        self.__dict__['_store']['min_time'] = ts_list[0]
+        self.__dict__['_store']['max_time'] = ts_list[-1]
+        self.__dict__['_store']['timestamps'] = ts_list
```
Jacob-Chmura, line 137: What if `edge_dict` is empty?

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+            start_time (int, optional): The start time of the aggregation. (default: :obj:`None`)
+            end_time (int, optional): The end time of the aggregation. (default: :obj:`None`)
+        Returns:
+            :class:`torch.Tensor`: The aggregated edge index tensor."""
+        if (start_time is None):
+            start_time = self.min_time
+        if (end_time is None):
+            end_time = self.max_time
+        assert start_time <= end_time, "start_time should be less or equal to end_time"
+        assert isinstance(start_time, int) and isinstance(end_time, int), "start_time and end_time should be integers"
```
Jacob-Chmura, line 217: Since they aren't being used to index anything, why do `start_time` and  `end_time` need to be integers? As long as they are numeric, this should be fine?

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+        out_index = self._store['events'][start_time].edges
+        for t in self._store['events'].keys():
+            if (t > start_time and t <= end_time):
```
Jacob-Chmura, line 220:
```suggestion
            if (start_time < t <= end_time):

```

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+        Returns:
+            :class:`torch.Tensor`: The aggregated edge index tensor."""
+        if (start_time is None):
+            start_time = self.min_time
+        if (end_time is None):
+            end_time = self.max_time
+        assert start_time <= end_time, "start_time should be less or equal to end_time"
+        assert isinstance(start_time, int) and isinstance(end_time, int), "start_time and end_time should be integers"
+        out_index = self._store['events'][start_time].edges
+        for t in self._store['events'].keys():
```
Jacob-Chmura, line 219: Minor, but we could iterate over our sorted list of timestamps and break early once we exceed `end_time`

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+            if (t > start_time and t <= end_time):
+                out_index = torch.cat((out_index, self._store['events'][t].edges), dim=1)
```
Jacob-Chmura, line 221: I have to think about the memory layout more, but generally, it is much faster to do a list appending in the inner loop and then just one single `torch.cat` on the final output

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+        return out_index
+
+    def to_events(self) -> Any:
+        r"""Converts a continuous time dynamic graph to a list of edge events.
+        Returns: a tuple of timestamps and edge index tensors."""
+        # Implement this method
+        edge_index = self._store['events'][self.min_time].edges
+        timestamps = [self.min_time]*(edge_index.shape[1])
+        for t in self._store['events'].keys():
+            if (t> self.min_time):
```
Jacob-Chmura, line 231: What's the reason for this?

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+                edge_index = torch.cat((edge_index, self._store['events'][t].edges), dim=1)
```
Jacob-Chmura, line 232: This could likely also be sped up by switching to a single `torch.cat` on the aggregated list of tensors.

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+                timestamps.extend([t]*(self._store['events'][t].edges.shape[1]))
+        timestamps = torch.tensor(timestamps)
+        return timestamps, edge_index
+    
+    def to_snapshots(self) -> Any:
+        """Converts a continuous time dynamic graph to a list of snapshots."""
+        # Implement this method
+        pass
+
+    def load_csv(self, file_path: str) -> Any:
```
Jacob-Chmura, line 242: I think it's not very clear why we would want someone to be able to `load_csv` on an already initiated `TG` object. This would basically just wipe out the previous data, shouldn't this just be a new object?

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+        """Loads a continuous time dynamic graph from a csv file."""
+        # Implement this method
+        edge_dict = self._load_csv(file_path)
+        self.__dict__['_store']['events'] = self._to_eventStore(edge_dict)
+
+    @property
+    def num_edges(self) -> Optional[int]:
+        r"""Returns the number of edges in the temporal graph."""
+        return sum([event.num_edges for event in self._store['events'].values()])
```
Jacob-Chmura, line 251: Should probably be cached

opendg/data/data.py
Jacob-Chmura, line 1: I understand this is a draft, but I left some initial comments below. At this point, it's worth splitting the different objects into separate files. Also, it's often the case that libraries will explicitly separate the interface and implementation (`CTDG`, `_CTDG_impl`), which allows them to test and swap implementation without making breaking changes to the API. It may be something worth considering since right now the two are tightly coupled

test/test_data/test_loading.py
Jacob-Chmura, line 1: I will leave this review for later, but it would be best to separate the tests into separate modules and handle the `setup()` and `tearDown()` for edge event data separately, test empty graphs, bad formats etc

opendg/data/storage.py
```diff
@@ -0,0 +1,362 @@
...
+    can contain edge or(and) node information"""
+    def __init__(
+        self,
+        timestamp: int = None,
+        edges: Optional[Tensor] = None,
+        node_feats: Optional[Dict[int, Tensor]] = None,
+        edge_feats: Optional[Tensor] = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__()
```
Jacob-Chmura, line 297: Should we not forward `timestamp` and `**kwargs` to the parent constructor?

opendg/data/storage.py
```diff
@@ -0,0 +1,362 @@
...
+        self._mapping: Dict[str, Any] = {}
```
Jacob-Chmura, line 298:
```suggestion
```

This is already in the parent class

opendg/data/storage.py
```diff
@@ -0,0 +1,362 @@
...
+        self._mapping['edges'] = edges
+        self._mapping['node_feats'] = node_feats
+        self._mapping['edge_feats'] = edge_feats
+        self._timestamp = timestamp
```
Jacob-Chmura, line 302:
```suggestion
```

This is already in the parent class (we should forward on the constructor)

Jacob-Chmura (commented):
Here is a preliminary review on what we have so far. The general structure makes some sense but I think there are a few areas that we should discuss in more detail

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+    Union,
+)
+from opendg.data.storage import EventStore
+from collections.abc import MutableMapping
+import csv
+import torch
+import os.path
+
+
+class BaseData:
```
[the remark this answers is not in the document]
shenyangHuang: yes should be abstract

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+    r"""Base class for all temporal graph data structures."""
+    def __getattr__(self, key: str) -> Any:
+        raise NotImplementedError
+
+    def __setattr__(self, key: str, value: Any):
+        raise NotImplementedError
+
+    def __delattr__(self, key: str):
+        raise NotImplementedError
```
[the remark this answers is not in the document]
shenyangHuang: to be discussed, happy to remove it as well, I followed base data structure from PyG as a starting point

opendg/data/data.py
```diff
@@ -0,0 +1,328 @@
...
+    def __copy__(self):
+        raise NotImplementedError
+
+    def __deepcopy__(self, memo):
+        raise NotImplementedError
+
+    def __repr__(self) -> str:
+        raise NotImplementedError
+    
+    def aggregate_graph(self, start_time:int, end_time:int) -> Any:
```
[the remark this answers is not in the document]
shenyangHuang: should not be DTDG but rather a static graph, though not sure how to represent it correctly, for now I only aggregate over edge_index, technicially we should aggregate over node features / have dynamic node features concatenated somehow as well

opendg/__init__.py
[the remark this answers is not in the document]
shenyangHuang: addressed

Jacob-Chmura:
Closing in light of https://github.com/shenyangHuang/openDG/pull/12

PR #12, opened 2024-12-18
DG Storage API

### LINKED ISSUE #13: DGStorage API

Jacob-Chmura:
Close #13 
### Purpose
The purpose of this PR is to add the `DGStorage` API, which is a private storage class that bundles the components of a dynamic graph. I have added a partially complete implementation of this API `DGStorageDict` which is not efficient but should serve as a good first MVP to test correctness. The `DGStorage` backends design will make it easy to test, benchmark, and swap out other implementations (e.g. tensor based) down the line.

### Outstanding Items
There are several incomplete components here that will be handled in subsequent PRs.

- [ ] Add proper `TimeDelta` object
- [ ] Figure out how to store node/edge features, and add them to API
- [ ] Integrate the `TimeDelta` object into `temporal_coarsening` and `time_granularity`
- [ ] Implement `temporal_coarsening` properly with the aggregation function

- get rid of nbr sampling from storage api
- add get_edge_index to storage API -- [time stamp array, 2 * |E|]


opendg/__init__.py
Jacob-Chmura, line 1: The `DGStorage` classes are intentionally excluded as they are intended to be used internally, not client-facing

opendg/storage/base.py
```diff
@@ -0,0 +1,74 @@
...
+    def materialize_edge_features(self) -> Tensor:
+        r"""Materialiize the dynamic graph edge feature data."""
+
+    @abstractmethod
+    def update(self, events: Union[Event, List[Event]]) -> 'DGStorageBase':
+        r"""Update the dynamic graph with an event of list of events."""
+
+    @abstractmethod
+    def temporal_coarsening(
+        self, time_delta: int, agg_func: str = 'sum'
```
Jacob-Chmura, line 47: This is a stub that will be expanded upon by creating a native `TimeDelta` type (#15 ) and properly annotating the set of possible aggregations, perhaps in an enum class.

opendg/storage/base.py
```diff
@@ -0,0 +1,74 @@
...
+    @abstractmethod
+    def slice_nodes(self, nodes: List[int]) -> 'DGStorageBase':
+        r"""Extract topological slice of the dynamcic graph given the list of nodes."""
+
+    @abstractmethod
+    def get_nbrs(self, nodes: List[int]) -> Dict[int, List[Tuple[int, int]]]:
+        r"""Return a list of neighbour, timestamp pairs for each node in the nodes list."""
+
+    @abstractmethod
+    def materialize_node_features(self) -> Tensor:
```
shenyangHuang, line 34: The size of the tensor would be $|T| \times |V| \times d$ where $|T|$ is # of timestamps, $|V|$ is the # of nodes, and $d$ is the dimension of the feature, we should add this in documentation

opendg/storage/base.py
```diff
@@ -0,0 +1,74 @@
...
+        r"""Materialiize the dynamic graph node feature data."""
+
+    @abstractmethod
+    def materialize_edge_features(self) -> Tensor:
```
shenyangHuang, line 38: $|T| \times |E| \times d$

opendg/storage/base.py
```diff
@@ -0,0 +1,74 @@
...
+
+from opendg.typing import Event
+
+
+class DGStorageBase(ABC):
+    r"""Base class for temporal graph storage engine."""
+
+    @classmethod
+    @abstractmethod
+    def from_events(cls, events: List[Event]) -> 'DGStorageBase':
```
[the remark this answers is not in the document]
shenyangHuang: sounds good, though it is expensive to create  / operate over large Lists in python. Base storage in PyG is `MutableMapping`, maybe we can change this to that as well. Dictionary are much more efficient for IO especially you don't need to know before hand the size (unlike lists)

opendg/storage/base.py
```diff
@@ -0,0 +1,74 @@
...
+    @abstractmethod
+    def materialize_node_features(self) -> Tensor:
+        r"""Materialiize the dynamic graph node feature data."""
+
+    @abstractmethod
+    def materialize_edge_features(self) -> Tensor:
+        r"""Materialiize the dynamic graph edge feature data."""
+
+    @abstractmethod
+    def update(self, events: Union[Event, List[Event]]) -> 'DGStorageBase':
```
shenyangHuang, line 42: This is more like appending new events to the end of the temporal graph, so I would say this can be called `append`? 

`update` could mean in the future that you can modify existing events if needed, currently we don't need to support that

opendg/storage/base.py
shenyangHuang: Yes I agree we can have a `TimeDelta` class to handle the conversion better, I propose a new way in the issue comment here #15

opendg/storage/base.py
```diff
@@ -0,0 +1,74 @@
...
+    def temporal_coarsening(
+        self, time_delta: int, agg_func: str = 'sum'
+    ) -> 'DGStorageBase':
+        r"""Re-index the temporal axis of the dynamic graph."""
+
+    @property
+    @abstractmethod
+    def start_time(self) -> int:
+        r"""The start time of the temporal graph."""
+
```
shenyangHuang, line 55:
```
@property
@abstractmethod
def time_granularity(self) -> TimeDelta: 
r"""Returns the specified time granularity of the temporal graph"""
```

opendg/typing.py
```diff
@@ -0,0 +1,3 @@
+from typing import Tuple
+
+Event = Tuple[int, int, int]  # Source Vertex, Target Vertex, TimeStamp
```
shenyangHuang, line 3: I would prefer the `Event` to be a class similar to my earlier implementation
In this way `Event` can have be both node event (add dynamic node feature) or edge event (add new edges + edge feature)

shenyangHuang (commented):
See discussions on typings and also #15

opendg/storage/base.py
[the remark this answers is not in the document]
Jacob-Chmura: Agreed. For smaller graphs, though, the cache locality might end up boosting the loop performance enough so that it doesn't matter much. I don't really have a sense of how much slower either is, so it would be nice to benchmark. I'm open to changing this when we implement the IO code.

opendg/storage/base.py
```diff
@@ -0,0 +1,74 @@
...
+    @abstractmethod
+    def slice_time(self, start_time: int, end_time: int) -> 'DGStorageBase':
+        r"""Extract temporal slice of the dynamic graph between start_time and end_time."""
+
+    @abstractmethod
+    def slice_nodes(self, nodes: List[int]) -> 'DGStorageBase':
+        r"""Extract topological slice of the dynamcic graph given the list of nodes."""
+
+    @abstractmethod
+    def get_nbrs(self, nodes: List[int]) -> Dict[int, List[Tuple[int, int]]]:
```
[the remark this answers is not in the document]
Jacob-Chmura: Yes that is how it would look like. To clarify, the alternative you suggested would return:
```
{u: {t1: (u, x), t2: (u, u)}, v: {t2: (v, y)}
```
?

opendg/storage/base.py
Jacob-Chmura: I have decided to remove the materialization from the public API since it is really an implementation detail. For instance, the `DGStorageDictBackend` is _always_ materialized, so this is mostly something that will be specific to the "sparse tensor" backend which does lazy ops.

opendg/storage/base.py
Jacob-Chmura: Good point, I will change this

opendg/storage/base.py
Jacob-Chmura: Thanks, yes this will be needed.

opendg/typing.py
Jacob-Chmura: I see, good point. I will think about reworking this.

opendg/storage/base.py
Jacob-Chmura: That being said, we do need a way to get the node / edge features

Issue #15, opened 2024-12-20
Add TimeDelta Object


Jacob-Chmura:
Idea here is to provide a utility class that encapsulates different time delta granularity levels: (`nanos`, `micros`, `millis`, `secs`, .., `days`, `weeks`, ...) and provides an API for re-indexing between different time deltas.

- Should also be easy to create  a custom time delta like: `TimeDelta("days", 5)`

Jacob-Chmura:
My suggestion is to adopt: https://numpy.org/doc/stable/reference/arrays.datetime.html

This makes temporal index conversions straightforward:

```python
# Example: nanos to micros
temporal_index = np.array([1, 2, 3, 10, 100, 1000], dtype='timedelta64[ns]')
temporal_index.astype('timedelta64[us]')

# Example: days to months
temporal_index = np.array([1, 2, 3, 10, 100, 1000], dtype='timedelta64[D]')
temporal_index.astype('timedelta64[Y]')
```

and also leverages fast kernels for index creation from timestamps.

shenyangHuang:
How fancy do we want this to be?

## Basic

close to what was implemented in TGX

- [ ] timestamps are integers 
- [ ] time granularity is specified with strings `year`, `month`, etc. 
- [ ] time granularity is specified for the `DGStorage`, one granularity accepted per TG 

### Pros: 

- simple and easy to use
- directly work with integers
- no new class needed

### Cons:
- timestamps are integer thus we need to make sure the correct time granularity exist for them
- constantly checking for valid strings etc. 

## `TimeDelta`

the above proposed implementation 

- [ ] timestamps all have their specific granularity as typing 
- [ ] conversion functions between time granularities directly on the array types
- [ ] adopt `numpy` Datetime64 typing

### Pros

- timestamps are enforced to have a time granularity
- the conversion function is intuitive and directly on the timestamps

### Cons
- what happens when a dataset has no time granularities (some TG datasets only provide the order of which edges arrive)
- How does it work on GPU, do we need to convert back to integer typing?

shenyangHuang:
Here is my idea of an in between solution

## `TimeDelta` Object Class (not bounded to timestamps) 

- `TimeDelta` class is initialized `TimeDelta("days", 5)` supports any string as time granularity and also duration as integer as input
- Timestamps remain as integer typing, easy to transfer between CPU and GPU
- a single `TimeDelta` class is required for each `DGStorage` (or temporal graph dataset)
[...]

PR #19, opened 2024-12-24
WIP: Dictionary Storage Backend

Jacob-Chmura:
Work in progress for _dictionary based_ `DGStorage` backend.


.pre-commit-config.yaml
Jacob-Chmura, line 1: I will merge this upstream since it's annoying to deal with otherwise.

test/test_storage_impl.py
Jacob-Chmura, line 1: This is a very crude test suite design. I want to expand on this and properly test across a broader range of graphs to catch edge cases.

test/test_storage_impl.py
```diff
@@ -0,0 +1,126 @@
...
+def test_update_multiple_events(DGStorageImpl):
+    events = []
+    storage = DGStorageImpl.from_events(events)
+    storage = storage.update([(1, 2, 3), (5, 10, 20)])
+    assert storage.to_events() == [(1, 2, 3), (5, 10, 20)]
+
+
+def test_get_dg_storage_backend():
+    backend = get_dg_storage_backend()
+    print(backend)
```
Jacob-Chmura, line 111:
``` suggestion
    assert get_dg_storage_backend() == DGStorageDictBackend
```

opendg/_storage/base.py
```diff
@@ -30,14 +28,6 @@ def slice_nodes(self, nodes: List[int]) -> 'DGStorageBase':
     def get_nbrs(self, nodes: List[int]) -> Dict[int, List[Tuple[int, int]]]:
         r"""Return a list of neighbour, timestamp pairs for each node in the nodes list."""
 
-    @abstractmethod
```
Jacob-Chmura, line 33: It's not immediately clear that we will need these as part of the generic API, since _laziness_ is an implementation detail more than anything. For instance, the current naive dictionary implementation is automatically materialized. Hence, I removed these for the time being.

opendg/_storage/base.py
```diff
@@ -48,14 +38,18 @@ def temporal_coarsening(
     ) -> 'DGStorageBase':
         r"""Re-index the temporal axis of the dynamic graph."""
 
+    def __len__(self) -> int:
```
Jacob-Chmura, line 41: Among other things, this seems relevant to have.

opendg/_storage/base.py
```diff
@@ -48,14 +38,18 @@ def temporal_coarsening(
...
     @property
     @abstractmethod
-    def start_time(self) -> int:
+    def start_time(self) -> Optional[int]:
         r"""The start time of the temporal graph."""
 
     @property
     @abstractmethod
-    def end_time(self) -> int:
+    def end_time(self) -> Optional[int]:
```
Jacob-Chmura, line 52: We need to decide upon certain graph semantics. I'm assuming we allow empty graphs. In that case, start/end time are undefined (`None`).

opendg/_storage/base.py
```diff
@@ -72,3 +66,9 @@ def num_edges(self) -> int:
     @abstractmethod
     def num_timestamps(self) -> int:
         r"""The total number of unique timestamps encountered over the temporal graph."""
+
+    def _check_slice_time_args(self, start_time: int, end_time: int) -> None:
```
Jacob-Chmura, line 70: Will continue to add in helper functions that will be common across backends into the base class (including device mixins, etc.)

opendg/_storage/backends/__init__.py
```diff
@@ -0,0 +1,7 @@
+from opendg._storage.backends.dict_backend import DGStorageDictBackend
+
+DGStorageBackends = {
+    'DictionaryBackend': DGStorageDictBackend,
```
Jacob-Chmura, line 4: Perhaps using the class name string itself is a bit better, I'm not entirely sure

opendg/_storage/backends/dict_backend.py
Jacob-Chmura, line 1: This is a very naive and incomplete implementation. There are plenty of optimizations to be had, and some methods can be written a bit better. But this is the gist of it which should enable benchmarks / correctness testing.

opendg/_storage/backends/dict_backend.py
```diff
@@ -0,0 +1,109 @@
...
+        if not isinstance(events, list):
+            events = [events]
+        for t, u, v in events:
+            self._events_dict[t] = (u, v)
+        return self
+
+    def temporal_coarsening(
+        self, time_delta: int, agg_func: str = 'sum'
+    ) -> 'DGStorageBase':
+        raise NotImplementedError()
```
Jacob-Chmura, line 72: TODO

opendg/_storage/backends/dict_backend.py
```diff
@@ -0,0 +1,109 @@
...
+        nbrs_dict = defaultdict(list)
+        for t, (u, v) in self._events_dict.items():
+            if u in nodes:
+                nbrs_dict[u].append((v, t))
+            if v in nodes:
+                nbrs_dict[v].append((u, t))
+
+        return dict(nbrs_dict)
+
+    def update(self, events: Union[Event, List[Event]]) -> 'DGStorageBase':
```
Jacob-Chmura, line 62: Need to update cache

opendg/_storage/backends/dict_backend.py
```diff
@@ -0,0 +1,110 @@
...
+from opendg.typing import Event
+
+from ..base import DGStorageBase
+
+
+class DGStorageDictBackend(DGStorageBase):
+    r"""Dictionary implementation of temporal graph storage engine."""
+
+    def __init__(self, events_dict: Dict[int, Tuple[int, int]]) -> None:
+        self._events_dict: Dict[int, Tuple[int, int]] = events_dict
```
Jacob-Chmura, line 13: Values should be `List[Tuple[int, int]]`

opendg/_storage/backends/dict_backend.py
```diff
@@ -0,0 +1,133 @@
...
+        events_dict = defaultdict(list)
+        for t, u, v in events:
+            events_dict[t].append((u, v))
+        return cls(events_dict)
+
+    def to_events(self) -> List[Event]:
+        events: List[Event] = []
+        for t, edges in self._events_dict.items():
+            for u, v in edges:
+                events.append((t, u, v))
```
shenyangHuang, line 33: this appending will be extremely slow for large graphs, 
this is one of the reasons earlier TG implementations are very slow and can't work on large graphs as they append to large lists

opendg/_storage/backends/dict_backend.py
```diff
@@ -0,0 +1,133 @@
...
+        self._end_time: Optional[int] = None
+        self._num_nodes: Optional[int] = None
+        self._num_edges: Optional[int] = None
+        self._num_timestamps: Optional[int] = None
+
+    @classmethod
+    def from_events(cls, events: List[Event]) -> 'DGStorageBase':
+        events_dict = defaultdict(list)
+        for t, u, v in events:
+            events_dict[t].append((u, v))
```
shenyangHuang, line 26: this should be fine for the most part, maybe slow for coarse time granlarity as you are appending many edges to a single timestamp

opendg/_storage/backends/dict_backend.py
```diff
@@ -0,0 +1,133 @@
...
+    ) -> 'DGStorageBase':
+        self._check_temporal_coarsening_args(time_delta, agg_func)
+
+        # This assert is here to make mypy happy
+        assert self.start_time is not None
+        assert self.end_time is not None
+        total_time = self.end_time - self.start_time
+
+        # TODO: Support other time_delta formats
+        interval_size = total_time // time_delta
```
shenyangHuang, line 84: #15 we can implement some similar idea here

opendg/_storage/base.py
shenyangHuang: Yes, this is the same as # of timestamps? or does this return some tensor sizes?

opendg/_storage/base.py
```diff
@@ -48,27 +38,45 @@ def temporal_coarsening(
...
+
+    def __str__(self) -> str:
+        r"""Returns summary properties of the dynamic graph."""
+        return f'Dynamic Graph Storage Engine ({self.__class__.__name__}), Start Time: {self.start_time}, End Time: {self.end_time}, Nodes: {self.num_nodes}, Edges: {self.num_edges}, Timestamps: {self.num_timestamps}'
+
     @property
     @abstractmethod
-    def start_time(self) -> int:
-        r"""The start time of the temporal graph."""
+    def start_time(self) -> Optional[int]:
```
shenyangHuang, line 51: why is it optional?

test/test_storage_impl.py
```diff
@@ -0,0 +1,179 @@
...
+    assert storage.end_time == 1
+    assert storage.num_nodes == 2
+    assert storage.num_edges == 1
+    assert storage.num_timestamps == 1
+
+
+def test_slice_time_bad_slice(DGStorageImpl):
+    events_dict = {1: (2, 3), 5: (10, 20)}
+    storage = DGStorageImpl(events_dict)
+    with pytest.raises(ValueError):
```
shenyangHuang, line 53: idea: maybe we should a custom class for errors? some errors would be invalid operation `InvalidOpsError` (like discretizing the graph into three snapshots when there are 2 edges only, need more thinking / testing). Others would be time related errors `ChronoError`, such as if you are slicing the timestamp with an interval either earlier than start time or later than end time?

test/test_storage_impl.py
```diff
@@ -0,0 +1,179 @@
...
+    pass
+
+
+@pytest.mark.skip(reason='Not implemented')
+def test_temporal_coarsening_single_event_graph(DGStorageImpl):
+    pass
+
+
+@pytest.mark.skip(reason='Not implemented')
+def test_temporal_coarsening_causes_time_gap(DGStorageImpl):
```
shenyangHuang, line 142: these should have specific error message telling the user this is no valid because xxx

opendg/_storage/base.py
Jacob-Chmura: Currently, I have implemented it as the # of timestamps, which I think is the most reasonable semantic for a dynamic graph _length_.

opendg/_storage/base.py
Jacob-Chmura: Because the graph could be empty

test/test_storage_impl.py
Jacob-Chmura: Yes, this is a good idea and we should do this whenever possible. For simplicity I've just raised ValueError's but having a custom error type for _time gaps_ makes sense. Generally though, I think using built in exception, where they make sense, is best (e.g. value error ["Raised when an operation or function receives an argument that has the right type but an inappropriate value"](https://docs.python.org/3/library/exceptions.html#ValueError))

opendg/_storage/backends/dict_backend.py
Jacob-Chmura: Can you elaborate? Perhaps I am missing something, but cpython lists are implemented as dynamic arrays. Therefore, appending to the end of a list has `O(1)` amortized time complexity (accounting for resizes), and has decent cache locality due to contiguous memory access (until the resize is triggered).

By your comment, do you simply mean that some appends will be much slower than others (when an array resize is triggered), or are you referring to the python loop which is, indeed very slow, but this does not really have much to do with the list appending?

PR #21, opened 2025-01-08
Add Features to DGStorage

Jacob-Chmura:


opendg/_storage/backends/dict_backend.py
```diff
@@ -1,17 +1,22 @@
...
 
 from ..base import DGStorageBase
 
 
 class DGStorageDictBackend(DGStorageBase):
     r"""Dictionary implementation of temporal graph storage engine."""
 
-    def __init__(self, events_dict: EventsDict) -> None:
-        self._events_dict: EventsDict = events_dict
+    def __init__(self, events: List[Event]) -> None:
```
Jacob-Chmura, line 16: I figured this list constructor is more relevant, easier to test, and closer to what we would expect after reading csv files, for example.

test/test_storage_impl.py
```diff
@@ -65,22 +30,26 @@ def test_init_from_events(DGStorageImpl):
...
+        EdgeEvent(time=1, edge=(10, 20)),
+        EdgeEvent(time=5, edge=(10, 20)),
+    ]
+    storage = DGStorageImpl(events)
     assert storage.to_events() == events
     assert storage.start_time == 1
     assert storage.end_time == 5
     assert storage.num_nodes == 4
-    assert storage.num_edges == 2
+    assert storage.num_edges == 3
```
Jacob-Chmura, line 44: We consider the second and third edges different since they occur at distinct time stamps.

test/test_storage_impl.py
```diff
@@ -14,73 +16,86 @@ def DGStorageImpl(request):
...
-    assert storage.to_events() == events
-    assert storage.start_time == 1
-    assert storage.end_time == 5
-    assert storage.num_nodes == 4
-    assert storage.num_edges == 2
-    assert storage.num_timestamps == 2
-    assert storage.time_granularity == 4
-    assert len(storage) == 2
 
+@pytest.mark.skip(reason='TODO: Handle incompatible event features')
```
Jacob-Chmura, line 85: It's not obvious to me how we should handle this, or if we are okay to assume that events passed into this internal object have been checked for consistency at the IO level

opendg/_storage/backends/dict_backend.py
```diff
@@ -138,6 +151,32 @@ def num_timestamps(self) -> int:
             self._num_timestamps = len(self._events_dict)
         return self._num_timestamps
 
+    @property
+    def node_feats(self) -> Optional[Tensor]:
```
Jacob-Chmura, line 155: Arguably this is worth caching

test/test_storage_impl.py
Jacob-Chmura, line 1: Apologies for the diff, but I kind of needed to add a `NodeEvent` to properly test this class, which changed a bunch of asserts

opendg/_storage/backends/dict_backend.py
```diff
@@ -138,6 +151,32 @@ def num_timestamps(self) -> int:
...
+                if isinstance(event, NodeEvent) and event.features is not None:
+                    feats.append(event.features)
+
+        if not len(feats):
+            return None
+
+        return torch.cat(feats)
+
+    @property
+    def edge_feats(self) -> Optional[Tensor]:
```
Jacob-Chmura, line 168: Question. Is this what we want? Or do we want a sparse Tensor of size `|T| x |V| x |V| x d_edge` where the missing feature coordinates are zero'd out? @shenyangHuang

(Analogous for node features)

Jacob-Chmura, line 168: How do we index the features

opendg/_storage/backends/dict_backend.py
```diff
@@ -21,18 +26,10 @@ def __init__(self, events_dict: EventsDict) -> None:
...
-            events_dict[t].append((u, v))
-        return cls(events_dict)
-
     def to_events(self) -> List[Event]:
         events: List[Event] = []
-        for t, edges in self._events_dict.items():
-            for u, v in edges:
-                events.append((t, u, v))
+        for t_events in self._events_dict.values():
+            events += t_events
```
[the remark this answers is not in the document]
shenyangHuang: or is this summing timestamps?

opendg/_storage/backends/dict_backend.py
shenyangHuang: yeah sparse tensor for indexing

shenyangHuang (approved):

opendg/_storage/backends/dict_backend.py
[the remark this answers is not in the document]
Jacob-Chmura: This is simply concatenating the list of events at the given timestamp. The events at the given timestamps are still _disjoint_, in the sense that the features are not concatenated.

opendg/_storage/backends/dict_backend.py
```diff
@@ -120,15 +146,25 @@ def time_granularity(self) -> Optional[TimeDelta]:
 
     @property
     def num_nodes(self) -> int:
-        if self._num_nodes is None:
-            edges = set(itertools.chain.from_iterable(self._events_dict.values()))
-            self._num_nodes = len(set(itertools.chain.from_iterable(edges)))
+        max_node_id = -1  # We assume the ids are >= 0
```
Jacob-Chmura, line 149: I have change the semantics so that the number of nodes in a DG is a function of the _max node id_ (assuming 0-based indexing) from all the events, rather than the number of individual node events. 

#### Example
- DG with a single `NodeEvent(time=3, node_id=5)`

**Previous**:
```python
DG.num_nodes == 1
```

**Now**
```python
DG.num_nodes == 6
```

PR #24, opened 2025-01-09
Implement basic event objects

### LINKED ISSUE #22: Create Event Object Hierarchy
- Node Event, Edge Event, etc.

Jacob-Chmura:
Keeping these straightforward and minimal for the time being.

Close #22 

Waiting for #12 to merge upstream first.


opendg/events.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self, time: int) -> None:
+        self._time = time
+
+    @property
+    def time(self) -> int:
+        return self._time
+
+
+class NodeEvent(Event):
```
Jacob-Chmura, line 18: Will consider creating a `TensorEvent` type that extends torch, so that we can automatically do things like move devices

Jacob-Chmura: Arguably these should be chronologically comparable (implementing __lt__, __gt__, etc), but until it's obvious we need this, I'll leave it out for now.

shenyangHuang: > Will consider creating a `TensorEvent` type that extends torch, so that we can automatically do things like move devices

I think that would be a good idea. 

< and > operations don't have to be on the event themselves, I agree.

shenyangHuang (approved):
look good!

PR #25, opened 2025-01-13
Adding TimeDelta Object

shenyangHuang:
- [x] First iteration of TimeDelta for openDG
- [x] test added for TimeDelta 
- [ ] implement time coarsening functionalities


opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
+from typing import Optional, Union
+
+# from opendg.typing import TimeDeltaTG
+
+
+class TimeDeltaTG:
```
Jacob-Chmura, line 6: Is there any reason we don't just call this `TimeDelta`?

opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
+from typing import Optional, Union
+
+# from opendg.typing import TimeDeltaTG
```
Jacob-Chmura, line 3:
```suggestion
```

opendg/typing.py
```diff
@@ -1,7 +1,8 @@
 from typing import Dict, List, Tuple
 
+from opendg.timedelta import TimeDeltaTG
+
```
Jacob-Chmura, line 4:
```suggestion
```

I don't think we really need this here. This file was primarily meant to contain type aliases of composite types (that are not an object).

opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class TimeDeltaTG:
+    r"""TimeGranularity class to represent time granularity in dynamic graph."""
+
+    def __init__(self, interval_type: str, duration: Optional[int] = 1):
```
Jacob-Chmura, line 9: This is mostly stylistic, but I would probably prefer matching the pandas/numpy parameter values, which are `value` and `unit` instead of `interval_type` and `duration`

opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
...
+        r"""Args:
+        interval_type (str): Type of time granularity. Possible values are 'Y', 'M', 'W', 'D', 'h', 'm', 's', 'ms', 'us', 'ns'.
```
Jacob-Chmura, line 11: We also allow `ordered` here?

opendg/timedelta.py
Jacob-Chmura, line 9:
```suggestion
    def __init__(self, interval_type: str, duration: int = 1):
```

Duration is not `Optional`, because it cannot be `None`.

opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
...
+            'ns': 'nanosecond',
+        }
+        self.non_convert = {'r': 'ordered'}
+
+        SECONDS_IN_A_YEAR = 31536000 #365 days
+        SECONDS_IN_A_MONTH = 2592000
+        SECONDS_IN_A_WEEK = 604800
+        SECONDS_IN_A_DAY = 86400
+        SECONDS_IN_AN_HOUR = 3600
+        SECONDS_IN_A_MINUTE = 60
```
Jacob-Chmura, line 34: I personally think it's a bit more intuitive if you just write out the terms in the multiplication to understand how you got the terms. E.g. `SECONDS_IN_A_DAY = SECONDS_IN_AN_HOUR * 24` or `SECONDS_IN_A_DAY = 60 * 60 * 24`.

opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
...
+    def interval_type(self) -> str:
+        r"""The time granularity type"""
+        return self.type
+
+    def _is_valid_type(self, type: str) -> bool:
+        r"""Check if the specified time granularity type is valid."""
+        return type in self.time_dict
+
+    def __str__(self):
+        return f'time granularity is {self.type} : {self.time_dict[self.type]}'
```
Jacob-Chmura, line 70: Wouldn't we want the `self.dur` to be reflected here? Something like `return f'{self.dur} {self.time_dict[self.type]}'` would do the trick

Jacob-Chmura: Well for `ordered` probably just `return "Ordered"`

opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __len__(self) -> int:
+        r"""Returns the number of duration of the specified time granularity."""
+        return self.dur
```
Jacob-Chmura, line 74: Personally I find this to be a counter-intuitive definition of length, mainly because it doesn't express the units. Can you give an example of why I would want to do something like `len(TimeDelta('M', 3))`? Also, the semantics become a bit confusing with `ordered` interval type

test/test_timedelta.py
Jacob-Chmura, line 1: This is a good start but we will definitely want to add more cases here. Especially with `ordered` type

opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
...
+        r"""Convert the time granularity to the specified time granularity type."""
+        new_secs = td.get_seconds()
+        cur_secs = self.get_seconds()
+        return cur_secs / new_secs
+
+    def get_seconds(self) -> float:
+        r"""Returns the time granularity in seconds."""
+        return self.dur * self.time_constant[self.type]
+
+    def convert(self, time_delta: Union[str, 'TimeDeltaTG']) -> int:
```
Jacob-Chmura, line 92: This is stylistic, but I usually like to keep public methods at the top of the class to see the API, and keep the private methods at the bottom.

opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
...
+            if time_delta.interval_type in self.non_convert:
+                raise ValueError(
+                    f'Conversion not allowed for time granularity type {time_delta.interval_type}'
+                )
+            else:
+                return self._convertFromDelta(time_delta)
+        else:
+            raise ValueError(
+                f'Invalid time granularity type for conversion: {time_delta}'
+            )
```
Jacob-Chmura, line 117: I think this can be somewhat cleaned up:
```suggestion
if isinstance(time_delta, str):
    time_delta = TimeDelta(time_delta)

if not isinstance(time_delta, TimeDelta):
    raise ValueError(f'Invalid type for conversion: {type(time_delta)}')

if type_delta.interval_type in self.non_convert:
    raise ValueError(f'Conversion not allowed for type {time_delta.interval_type}')

return self._convertFromDelta(time_delta)
```

opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
...
+        r"""Returns the number of duration of the specified time granularity."""
+        return self.dur
+
+    def _convertFromStr(self, interval_type: str, duration: int) -> float:
+        r"""Convert the time granularity to the specified time granularity type."""
+        new_secs = duration * self.time_constant[interval_type]
+        cur_secs = self.get_seconds()
+        return cur_secs / new_secs
+
+    def _convertFromDelta(self, td: 'TimeDeltaTG') -> float:
```
Jacob-Chmura, line 82: Stylistic
```suggestion
    def _convert_from_timedelta(self, td: 'TimeDeltaTG') -> float:
```

opendg/timedelta.py
```diff
@@ -0,0 +1,117 @@
...
+            'Y': SECONDS_IN_A_YEAR,
+            'M': SECONDS_IN_A_MONTH,
+            'W': SECONDS_IN_A_WEEK,
+            'D': SECONDS_IN_A_DAY,
+            'h': SECONDS_IN_AN_HOUR,
+            'm': SECONDS_IN_A_MINUTE,
+            's': 1,
+            'ms': 1 / 1000,
+            'us': 1 / 1000000,
+            'ns': 1 / 1000000000,
```
Jacob-Chmura, line 45:
```suggestion
            'us': 1 / 1_000_000,
            'ns': 1 / 1_000_000_000,
```

opendg/timedelta.py
Jacob-Chmura: Also, this returns a float.

opendg/timedelta.py
Jacob-Chmura, line 1: Is it worth adding ordering semantics to enable things like `TimeDelta('M', 3) < TimeDelta('Y')`, `TimeDelta(`h`, 2) == TimeDelta(`m`, 120)`?

opendg/timedelta.py
Jacob-Chmura: Personally, `convert` makes it sound like we are _converting_ the `self` time delta into the interval_type of the parameter time_delta, whereas we are actually computing the _quotient_. Don't really have a better name of the top off my head, but maybe it will be more clear when we use this object to implement the coarsening ops.

opendg/timedelta.py
shenyangHuang: I think Python has a native `TimeDelta`? also numpy also called their `TimeDelta64` so I want to make a distinction here. In addition, our time delta is only for conversion between time granularities, but not necessarily for between two real world dates as of now

opendg/typing.py
shenyangHuang: agree

opendg/timedelta.py
shenyangHuang: agree, fixed

opendg/timedelta.py
shenyangHuang: yes, `r` means ordered

opendg/timedelta.py
shenyangHuang: agree

opendg/timedelta.py
shenyangHuang: potentially, though you really just need the convert function for the majority of cases

PR #27, opened 2025-01-26
IO Implementation

Jacob-Chmura:
WIP


opendg/_io/csv.py
Jacob-Chmura, line 1: Need to discuss some implementation details here, especially around how we infer column types when reading and writing data, how we deal with multi-dimensional arrays, etc

opendg/graph.py
```diff
@@ -0,0 +1,104 @@
+from typing import Any, List, Optional, Union
+
+from opendg._events import Event
+from opendg._io import read_csv, write_csv
+from opendg._storage import DGStorage
+from opendg.typing import TimeDelta
+
+
+class DGraph:
```
Jacob-Chmura, line 9: Need to discuss this further, but this would evolve into the user facing API that clients actually use (as opposed to the storage). If we continue down this route, we need to be clear around what is / needs to be supported directly for the user

test/test_io/test_csv_io.py
Jacob-Chmura, line 1: There are some fixtures that I will create that exhausts the different cases in more detail. This is just a starting point

opendg/_io/csv.py
```diff
@@ -0,0 +1,87 @@
...
+    events: List[Event] = []
+    with open(file_path, newline='') as f:
+        reader = csv.DictReader(f)
+        for row in reader:
+            src_id = int(row[src_node_id_col])
+            dst_id = int(row[dst_node_id_col])
+            time = int(row[time_col])
+
+            if feature_cols is None:
+                features = None  # TODO:: Infer the feature columns
```
Jacob-Chmura, line 26: Will circle back to this when we start putting in `examples` and `benchmarks`. Then it will become more clear how the inference should be best structured.

PR #37, opened 2025-02-02
Graph Sampling

### LINKED ISSUE #36: Recency Neighbour Sampling
[Andy] adding some ideas here:

- [ ] recency neighbor sampler should inherent from `BaseLoader`
- [ ] recent neighbor sampler keep tracks of a $|\mathbf{N}| \times k$ matrix which stores $k$ neighbors for each nodes in DGraph ($k$ is a user input parameter at initialization)

Jacob-Chmura:
Close #36

### DIFF
--- a/opendg/sampler/base.py
+++ b/opendg/sampler/base.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Any, List, Optional
+from typing import List, Optional
 
 from opendg.graph import DGraph
 
@@ -13,7 +13,6 @@ class BaseDGSampler(ABC):
         seed_nodes: Optional[List[int]] = None,
         from_timestamp: Optional[int] = None,
         until_timestamp: Optional[int] = None,
-        *kwargs: Any,
     ) -> DGraph:
         r"""Sample seed nodes from a dynamic graph in the temporal window [from_timestamp, until_timestamp].
 
@@ -21,7 +20,6 @@ class BaseDGSampler(ABC):
             seed_nodes (Optional[List[int]]): The indices of the seed nodes to sample from, or the entire graph if None.
             from_timestamp (Optional[int]): The minimum time to sampling from, or the dynamic graph start time if None.
             until_timestamp (Optional[int]): The maximum time to sampling until, or the dynamic graph start time if None.
-            kwargs (Any): Optional sampler specific keyword arguments.
 
         Returns:
             A subgraph view on the DGraph according to the sampling strategy and input arguments.
@@ -29,7 +27,7 @@ class BaseDGSampler(ABC):
 
     @property
     @abstractmethod
-    def num_neighbours(self) -> List[int]:
+    def num_neighbors(self) -> List[int]:
         r"""The number of neighbours to sampler for each hop."""
 
     @property
--- a/opendg/sampler/base.py
+++ b/opendg/sampler/base.py
@@ -1,0 +1,41 @@
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional
+
+from opendg.graph import DGraph
+
+
+class BaseDGSampler(ABC):
+    r"""Base class for sampling from a dynamic graph."""
+
+    @abstractmethod
+    def sample(
+        self,
+        seed_nodes: Optional[List[int]] = None,
+        from_timestamp: Optional[int] = None,
+        until_timestamp: Optional[int] = None,
+        *kwargs: Any,
+    ) -> DGraph:
+        r"""Sample seed nodes from a dynamic graph in the temporal window [from_timestamp, until_timestamp].
+
+        Args:
+            seed_nodes (Optional[List[int]]): The indices of the seed nodes to sample from, or the entire graph if None.
+            from_timestamp (Optional[int]): The minimum time to sampling from, or the dynamic graph start time if None.
+            until_timestamp (Optional[int]): The maximum time to sampling until, or the dynamic graph start time if None.
+            kwargs (Any): Optional sampler specific keyword arguments.
+
+        Returns:
+            A subgraph view on the DGraph according to the sampling strategy and input arguments.
+        """
+
+    @property
+    @abstractmethod
+    def num_neighbours(self) -> List[int]:
+        r"""The number of neighbours to sampler for each hop."""
+
+    @property
+    @abstractmethod
+    def num_hops(self) -> int:
+        r"""The number of hops specified in the sampler."""
+
+    def __len__(self) -> int:
+        return self.num_hops
--- a/opendg/sampler/neighbor_sampler.py
+++ b/opendg/sampler/neighbor_sampler.py
@@ -1,0 +1,63 @@
+from typing import Any, Callable, List, Optional, Union
+
+from opendg.graph import DGraph
+from opendg.sampler.base import BaseDGSampler
+
+from .base import BaseDGSampler
+from .sampling_func import SamplingFunc, construct_sampling_func
+
+SamplingFuncType = Union[
+    str,
+    SamplingFunc,
+    Callable[[float], float],
+]
+
+
+class DGNeighborSampler(BaseDGSampler):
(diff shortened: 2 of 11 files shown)


opendg/sampler/neighbor_sampler.py
```diff
@@ -0,0 +1,63 @@
...
+        self._num_neighbors = num_neighbors
+        self._sampling_func = construct_sampling_func(sampling_func, **kwargs)
+
+    def sample(
+        self,
+        seed_nodes: Optional[List[int]] = None,
+        from_timestamp: Optional[int] = None,
+        until_timestamp: Optional[int] = None,
+    ) -> DGraph:
+        # TODO
```
Jacob-Chmura, line 49: Basically, this will look like:
```python
_dg = self._dg.create_view() # Create a materialized view
_dg = _dg.slice_time(from_timestamp, until_timestamp) # if the timestamps are not none
for i in range(self.num_hop):
    _dg = _dg.slice_nodes(seed_nodes)
    temporal_weights = self._sampling_func(_dg.timestamps)
    temporal_samples = sample(temporal_weights)
    seed_nodes = disjoint_union(seed_nodes, temporal_samples)
return _dg

```
1. Slice the entire graph temporally if the timestamps are given
2. For each hop in `self.num_hops`, slice nodes on seed nodes. Appl

opendg/sampler/sampling_func.py
```diff
@@ -0,0 +1,95 @@
...
+    if arg in _sampling_funcs:
+        sampling_func_klass = _sampling_funcs[arg]
+        return sampling_func_klass(**kwargs)
+    else:
+        raise ValueError(
+            f'Unknown sampling function type: {arg}, use one of: {list(_sampling_funcs.keys())}'
+        )
+
+
+class SamplingFunc(ABC):
```
Jacob-Chmura, line 26: Note: if we want to do something like "keep the last k neighbors", then we actually need access to the underlying store / timestamp info of the dg view.

opendg/sampler/neighbor_sampler.py
Jacob-Chmura: And also some re-indexing logic here

opendg/sampler/base.py
```diff
@@ -0,0 +1,39 @@
...
+        seed_nodes: Optional[List[int]] = None,
+        from_timestamp: Optional[int] = None,
+        until_timestamp: Optional[int] = None,
+    ) -> DGraph:
+        r"""Sample seed nodes from a dynamic graph in the temporal window [from_timestamp, until_timestamp].
+
+        Args:
+            seed_nodes (Optional[List[int]]): The indices of the seed nodes to sample from, or the entire graph if None.
+            from_timestamp (Optional[int]): The minimum time to sampling from, or the dynamic graph start time if None.
+            until_timestamp (Optional[int]): The maximum time to sampling until, or the dynamic graph start time if None.
```
shenyangHuang, line 22: or the dynamic graph end time if None right.

opendg/sampler/neighbor_sampler.py
```diff
@@ -0,0 +1,63 @@
...
+        dg (DGraph): The original dynamic graph to sample from.
+        num_neighbors (List[int]): The number of neighbors to sample at each hop.
+        sampling_func (Union[str, SamplingFunc, Callable[[float], float]]): The temporal sampling function to apply.
+    """
+
+    def __init__(
+        self,
+        dg: DGraph,
+        num_neighbors: Union[int, List[int]],
+        sampling_func: SamplingFuncType,
```
shenyangHuang, line 29: here it should be `(Union[str, SamplingFunc, Callable[[float], float]])` too right similar to the documentation? what is `Callable[[float], float]]`?

opendg/sampler/neighbor_sampler.py
shenyangHuang: Do we need to make a specific `create_view()` function before calling any of `slice_time` or `slice_nodes` ?

opendg/sampler/recency_neighbor_sampler.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any, List, Union
+
+from opendg.graph import DGraph
+
+from .neighbor_sampler import DGNeighborSampler
+
+
+class DGRecencyNeighborSampler(DGNeighborSampler):
```
shenyangHuang, line 8: we can implement this more later as it is less used, however, should there be a cutoff threshold?

opendg/sampler/sampling_func.py
```diff
@@ -0,0 +1,95 @@
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Optional, Union
+
+import numpy as np
+
+
+def construct_sampling_func(
+    arg: Union[str, Callable[[float], float], 'SamplingFunc'],
```
shenyangHuang, line 8: still not sure what is `Callable[[float], float]`? if it is to describe # of neighbors similar to pyg, it would be List[[int]]?

opendg/sampler/sampling_func.py
shenyangHuang: yes, actually keeping the last k neighbors is very efficient because you would keep a local stash, or any data structure that is first in first out, and push and pop each time new data arrives?

test/test_seed.py
```diff
@@ -0,0 +1,17 @@
+import random
+
+import numpy as np
+import torch
+
+from opendg.seed import seed_everything
+
+
+def test_seed_everything():
```
shenyangHuang, line 9: this looks good, I would add more tests once we have layers to test if we can reproduce from the same seed

shenyangHuang (commented):
Looks good overall, let's discuss more on design choices and structure

opendg/sampler/sampling_func.py
Jacob-Chmura: Yeah, makes me think we need to have an extra layer in this hierarchy:
```
SamplingFunc: something really basic and general which has access to the underlying dg

TemporalSamplingFunc: only cares about timestamp sequence (what is currently `SamplingFunc` above)
```

opendg/sampler/base.py
Jacob-Chmura: Correct, will fix

opendg/sampler/neighbor_sampler.py
Jacob-Chmura: `SamplingFuncType` is just a type alias for `(Union[str, SamplingFunc, Callable[[float], float]])`. They mean the same thing and can be used interchangeably.

`Callable[[float], float]]` is the type signature of a function that take a single `float` as an input argument and produces a single `float` as an output. In this context, the idea is that it is some function which determines the sampling weights as a function of time change. For instance, a decaying exponential means your sampling probability drops exponential for distant temporal neighbors, a uniform sampling probability means you sample uniformly across time (up to some cutoff).

Jacob-Chmura: This signature is meant to make it easy to do each of the three things, respectively:

```python
Sampler(dg, num_neighbors=5, sampling_func='uniform') # use uniform sampling

Sampler(dg, num_neighbors=5, sampling_func=UniformSamplingFunc) # use the build-in uniform sampling class

def sampling_func(dt):
    return 0 if dt > 5 else 1

Sampler(dg, num_neighbors=5, sampling_func=sampling_func) # use custom lambda or function
```

opendg/sampler/recency_neighbor_sampler.py
Jacob-Chmura: Honestly, if it's not top priority, I will just remove it to keep things simple for now. But yes, I have included a cutoff threshold: https://github.com/shenyangHuang/openDG/blob/640f92f3c3edf352612651d05815e6a3fd40233d/opendg/sampler/sampling_func.py#L73-L76

opendg/sampler/sampling_func.py
Jacob-Chmura: See previous comment

PR #51, opened 2025-02-26
Adding EdgeBank with simple test set 

shenyangHuang:
- added Edgebank in `nn/memory/edgebank.py`
- added simple test example for edgebank

### DIFF
--- a/opendg/nn/memory/edgebank.py
+++ b/opendg/nn/memory/edgebank.py
@@ -26,7 +26,7 @@ class EdgeBankPredictor:
             pos_prob(float): the probability of the link existence for the edges in memory.
         """
         if memory_mode not in ['unlimited', 'fixed_time_window']:
-            raise ValueError(f'Invalid memory mode for EdgeBank, expected one of {...}')
+            raise ValueError(f'Invalid memory mode for EdgeBank, expected one of ["unlimited", "fixed"]')
 
         if (time_window_ratio <= 0.0) or (time_window_ratio > 1.0):
             raise ValueError(
--- a/opendg/nn/memory/edgebank.py
+++ b/opendg/nn/memory/edgebank.py
@@ -1,9 +1,7 @@
-import warnings
-
 import numpy as np
 
 
-class EdgeBankPredictor(object):
+class EdgeBankPredictor:
     def __init__(
         self,
         src: np.ndarray,
@@ -12,35 +10,38 @@ class EdgeBankPredictor(object):
         memory_mode: str = 'unlimited',  # could be `unlimited` or `fixed_time_window`
         time_window_ratio: float = 0.15,
         pos_prob: float = 1.0,
-    ):
+    ) -> None:
         r"""Intialize edgebank and specify the memory mode.
-        Parameters:
-            src: source node id of the edges for initialization
-            dst: destination node id of the edges for initialization
-            ts: timestamp of the edges for initialization
-            memory_mode: 'unlimited' or 'fixed_time_window'
-            time_window_ratio: the ratio of the time window length to the total time length
-            pos_prob: the probability of the link existence for the edges in memory.
+
+        Args:
+            src(np.ndarray): source node id of the edges for initialization
+            dst(np.ndarray): destination node id of the edges for initialization
+            ts(np.ndarray): timestamp of the edges for initialization
+            memory_mode(str): 'unlimited' or 'fixed_time_window'
+            time_window_ratio(float): the ratio of the time window length to the total time length
+            pos_prob(float): the probability of the link existence for the edges in memory.
         """
-        assert memory_mode in [
-            'unlimited',
-            'fixed_time_window',
-        ], 'Invalide memory mode for EdgeBank!'
+        if memory_mode not in ['unlimited', 'fixed_time_window']:
+            raise ValueError(f'Invalide memory mode for EdgeBank')
+
+        if (time_window_ratio <= 0.0) or (time_window_ratio > 1.0):
+            raise ValueError(
+                f'Invalide time window ratio for EdgeBank, must be in (0,1]'
+            )
+
         self.memory_mode = memory_mode
+
+        self.cur_t = ts.max()
+
         if self.memory_mode == 'fixed_time_window':
-            self.time_window_ratio = time_window_ratio
             # determine the time window size based on ratio from the given src, dst, and ts for initialization
             duration = ts.max() - ts.min()
             self.prev_t = ts.min() + duration * (
-                1 - time_window_ratio
+                1.0 - time_window_ratio
             )  # the time windows starts from the last ratio% of time
-            self.cur_t = ts.max()
-            self.duration = self.cur_t - self.prev_t
         else:
-            self.time_window_ratio = -1
-            self.prev_t = -1
-            self.cur_t = -1
-            self.duration = -1
+            self.prev_t = ts.min()
+        self.duration = self.cur_t - self.prev_t
 
         self.memory: dict[tuple[int, int], int] = {}  # {(u,v):1}
         self.pos_prob = pos_prob
@@ -49,10 +50,11 @@ class EdgeBankPredictor(object):
     def update_memory(self, src: np.ndarray, dst: np.ndarray, ts: np.ndarray) -> None:
         r"""Generate the current and correct state of the memory with the observed edges so far.
         note that historical edges may include training, validation, and already observed test edges.
-        Parameters:
-            src: source node id of the edges.
-            dst: destination node id of the edges.
-            ts: timestamp of the edges.
+
+        Args:
+            src(np.ndarray): source node id of the edges.
+            dst(np.ndarray): destination node id of the edges.
+            ts(np.ndarray): timestamp of the edges.
         """
         if self.memory_mode == 'unlimited':
             self._update_unlimited_memory(src, dst)  # ignores time
@@ -63,37 +65,30 @@ class EdgeBankPredictor(object):
 
     @property
     def start_time(self) -> int:
-        """Return the start of time window for edgebank `fixed_time_window` only.
+        r"""The start of time window for edgebank.
 
         Returns:
-            start of time window.
+            int: start of time window.
         """
-        if self.memory_mode == 'unlimited':
-            warnings.warn(
-                'start_time is not defined for unlimited memory mode, returns -1'
-            )
         return self.prev_t
 
     @property
     def end_time(self) -> int:
-        """Return the end of time window for edgebank `fixed_time_window` only.
+        r"""The end of time window for edgebank.
 
         Returns:
-            end of time window.
+            int: end of time window.
         """
-        if self.memory_mode == 'unlimited':
-            warnings.warn(
-                'end_time is not defined for unlimited memory mode, returns -1'
-            )
         return self.cur_t
 
     def _update_unlimited_memory(
         self, update_src: np.ndarray, update_dst: np.ndarray
     ) -> None:
         r"""Update self.memory with newly arrived src and dst.
-        Parameters:
-            src: source node id of the edges.
-            dst: destination node id of the edges.
+
+        Args:
+            update_src(np.ndarray): source node id of the edges.
+            update_dst(np.ndarray): destination node id of the edges.
         """
         for src, dst in zip(update_src, update_dst):
             if (src, dst) not in self.memory:
@@ -104,17 +99,12 @@ class EdgeBankPredictor(object):
     ) -> None:
         r"""Move the time window forward until end of dst timestamp here.
         also need to remove earlier edges from memory which is not in the time window.
-        Parameters:
-            update_src: source node id of the edges.
-            update_dst: destination node id of the edges.
-            update_ts: timestamp of the edges.
-        """
-        # * initialize the memory if it is empty
-        if len(self.memory) == 0:
-            for src, dst, ts in zip(update_src, update_dst, update_ts):
-                self.memory[(src, dst)] = ts
-            return None
 
+        Args:
+            update_src(np.ndarray): source node id of the edges.
+            update_dst(np.ndarray): destination node id of the edges.
+            update_ts(np.ndarray): timestamp of the edges.
+        """
         # * update the memory if it is not empty
         if update_ts.max() > self.cur_t:
             self.cur_t = update_ts.max()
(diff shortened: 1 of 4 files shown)


opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
+import warnings
+
+import numpy as np
+
+
+class EdgeBankPredictor(object):
```
Jacob-Chmura, line 6:
```suggestion
class EdgeBankPredictor:
```

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+            dst: destination node id of the edges for initialization
+            ts: timestamp of the edges for initialization
+            memory_mode: 'unlimited' or 'fixed_time_window'
+            time_window_ratio: the ratio of the time window length to the total time length
+            pos_prob: the probability of the link existence for the edges in memory.
+        """
+        assert memory_mode in [
+            'unlimited',
+            'fixed_time_window',
+        ], 'Invalide memory mode for EdgeBank!'
```
Jacob-Chmura, line 28: Nit, but I suggest all exceptions of this sort in public API's raise something more specific, like `ValueError`, rather than implicit `AssertionError`. Technically assertions are not guaranteed to run, and are meant more as debug checks.

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+            self.prev_t = ts.min() + duration * (
+                1 - time_window_ratio
+            )  # the time windows starts from the last ratio% of time
+            self.cur_t = ts.max()
+            self.duration = self.cur_t - self.prev_t
+        else:
+            self.time_window_ratio = -1
+            self.prev_t = -1
+            self.cur_t = -1
+            self.duration = -1
```
Jacob-Chmura, line 43: Somehow I feel that this can be consolidated with the _finite window_ setup. For instance

```python
self.prev_t = ts.min()
self.cur_t = ts.max()
self.duration = self.cur_t - self.prev_t
```
all seem valid.

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+
+import numpy as np
+
+
+class EdgeBankPredictor(object):
+    def __init__(
+        self,
+        src: np.ndarray,
+        dst: np.ndarray,
+        ts: np.ndarray,
```
Jacob-Chmura, line 11: We don't allow the user to start with empty data?

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+        else:
+            raise ValueError('Invalide memory mode!')
+
+    @property
+    def start_time(self) -> int:
+        """Return the start of time window for edgebank `fixed_time_window` only.
+
+        Returns:
+            start of time window.
+        """
```
Jacob-Chmura, line 70: Can we use r-strings everywhere for consistency, and add types to the arguments and returns on all public API's, since these will be shown in our documentation. Also, note that all private methods (begin with underscores) are not included in docs, so we don't necessarily need to document them.

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+            self.time_window_ratio = -1
+            self.prev_t = -1
+            self.cur_t = -1
+            self.duration = -1
+
+        self.memory: dict[tuple[int, int], int] = {}  # {(u,v):1}
+        self.pos_prob = pos_prob
+        self.update_memory(src, dst, ts)
+
+    def update_memory(self, src: np.ndarray, dst: np.ndarray, ts: np.ndarray) -> None:
```
Jacob-Chmura, line 49: Is the intention that these arrays are produced by our data loader? Seems to me it would be easier to just update from a `DGraph` but I could be missing something

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+            update_ts: timestamp of the edges.
+        """
+        # * initialize the memory if it is empty
+        if len(self.memory) == 0:
+            for src, dst, ts in zip(update_src, update_dst, update_ts):
+                self.memory[(src, dst)] = ts
+            return None
+
+        # * update the memory if it is not empty
+        if update_ts.max() > self.cur_t:
```
Jacob-Chmura, line 119: This `.max()` will throw a `ValueError` if the array is empty. We should check for emptiness and make the update a no-op.

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+        r"""Move the time window forward until end of dst timestamp here.
+        also need to remove earlier edges from memory which is not in the time window.
+        Parameters:
+            update_src: source node id of the edges.
+            update_dst: destination node id of the edges.
+            update_ts: timestamp of the edges.
+        """
+        # * initialize the memory if it is empty
+        if len(self.memory) == 0:
+            for src, dst, ts in zip(update_src, update_dst, update_ts):
```
Jacob-Chmura, line 114: As in my previous comment, I'm not sure if we want to expose updates over `np.ndarray` vs. `DGraph`, but if we are doing it over arrays, we should probably throw a `ValueError` or a warning if the shapes of the arrays don't match, since this `zip` will truncate silently to the shortest array

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+                self.memory[(src, dst)] = ts
+            return None
+
+        # * update the memory if it is not empty
+        if update_ts.max() > self.cur_t:
+            self.cur_t = update_ts.max()
+            self.prev_t = self.cur_t - self.duration
```
Jacob-Chmura, line 121: Shouldn't we remove stale edges from the bank? I would assume this is one of the reasons someone might choose a fixed window instead of unlimited (the other being that they don't want really old edges to influence current prediction)

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+            )  # the time windows starts from the last ratio% of time
+            self.cur_t = ts.max()
+            self.duration = self.cur_t - self.prev_t
+        else:
+            self.time_window_ratio = -1
+            self.prev_t = -1
+            self.cur_t = -1
+            self.duration = -1
+
+        self.memory: dict[tuple[int, int], int] = {}  # {(u,v):1}
```
Jacob-Chmura, line 45: I might be missing something, but what are we using a map for? Do we ever access the values? I suspect we can just use a set.

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+            query_dst: destination node id of the query edges.
+
+        Returns:
+            pred: the prediction for all query edges.
+        """
+        pred = np.zeros(len(query_src))
+        idx = 0
+        for src, dst in zip(query_src, query_dst):
+            if (src, dst) in self.memory:
+                if self.memory_mode == 'fixed_time_window':
```
Jacob-Chmura, line 141: If we make self.prev_t = INT_MIN or `ts.min()` etc in for _unlimited_memory_ edgebank then we wouldn't have to handle these seperately, since `self.memory[(src, dst)]  >= self.prev_t` would just always be true for unlimited.

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+        Parameters:
+            update_src: source node id of the edges.
+            update_dst: destination node id of the edges.
+            update_ts: timestamp of the edges.
+        """
+        # * initialize the memory if it is empty
+        if len(self.memory) == 0:
+            for src, dst, ts in zip(update_src, update_dst, update_ts):
+                self.memory[(src, dst)] = ts
+            return None
```
Jacob-Chmura, line 116: Why do we need this special case? As far as I can tell, we can just remove this block and everything should still work even if `self.memory` is empty, so long as `self.cur_t = INT_MIN`. But also, the current constructor implementation breaks if we try initializing with empty arrays anyway

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,148 @@
...
+
+class EdgeBankPredictor(object):
+    def __init__(
+        self,
+        src: np.ndarray,
+        dst: np.ndarray,
+        ts: np.ndarray,
+        memory_mode: str = 'unlimited',  # could be `unlimited` or `fixed_time_window`
+        time_window_ratio: float = 0.15,
+        pos_prob: float = 1.0,
```
Jacob-Chmura, line 14: I would at least argument check the `time_window_ratio` to ensure it is in `[0, 1]` since otherwise it could lead to issue where `self.prev_t > self.curr_t`

test/test_model/test_edgebank.py
```diff
@@ -0,0 +1,50 @@
...
+
+def test_no_loader(DGStorageImpl):
+    events = [
+        EdgeEvent(time=1, edge=(2, 3), features=torch.rand(2, 5)),
+        EdgeEvent(time=5, edge=(10, 20), features=torch.rand(2, 5)),
+        NodeEvent(time=6, node_id=7, features=torch.rand(3, 6)),
+    ]
+    td = TimeDeltaDG('Y', 1)
+    storage = DGStorageImpl(events, td)
+    events = storage.to_events()
```
Jacob-Chmura, line 24: What's the point of going back and forth to and from events?

test/test_model/test_edgebank.py
```diff
@@ -0,0 +1,50 @@
...
+    for event in events:
+        if isinstance(event, EdgeEvent):
+            edges.append([event.edge[0], event.edge[1], event.time])
+    edge_index = np.array(edges)
+    srcs = edge_index[:, 0]
+    dsts = edge_index[:, 1]
+    ts = edge_index[:, 2]
+
+    MEMORY_MODE = 'unlimited'
+    TIME_WINDOW_RATIO = 0.15
```
Jacob-Chmura, line 35: Since the time window is ignored under 'unlimited' memory, we should probably add a test that actually uses the time window, does updates, and checks edges that _would be there_ but are beyond our memory are _ignored_.

And also that the `start_time, end_time` properties are updated correctly, since we have added them to the api

Jacob-Chmura (changes requested):

opendg/nn/memory/edgebank.py
shenyangHuang: updated

opendg/nn/memory/edgebank.py
shenyangHuang: I agree, simplified.

opendg/nn/memory/edgebank.py
shenyangHuang: this can't be empty for fixed time window because the time window ratio is dependent on timestamps from the dataset

opendg/nn/memory/edgebank.py
shenyangHuang: yes got it

shenyangHuang: should we remove documentation for private methods or is it fine to keep them?

opendg/nn/memory/edgebank.py
shenyangHuang: yeah we don't need it

opendg/nn/memory/edgebank.py
shenyangHuang: yes, added check, actually we will enforce `(0,1]`, otherwise there is no point of using Edgebank if ratio=0

test/test_model/test_edgebank.py
shenyangHuang: was trying to see if we can get data from storage, but this is unnecessary to test here

test/test_model/test_edgebank.py
shenyangHuang: added test for fixed time window

shenyangHuang:
Edgebank is ready for merge now.

opendg/nn/memory/edgebank.py
```diff
@@ -0,0 +1,125 @@
+import numpy as np
+
+
+class EdgeBankPredictor:
+    def __init__(
+        self,
+        src: np.ndarray,
+        dst: np.ndarray,
+        ts: np.ndarray,
+        memory_mode: str = 'unlimited',  # could be `unlimited` or `fixed_time_window`
```
Jacob-Chmura, line 10: We may want to consider the following type to enforce this:
```python
memory_mode: Literal['unlimited', 'fixed_time_window'] = 'unlimited',
```

test/test_model/test_edgebank.py
```diff
@@ -0,0 +1,84 @@
...
+    edges = []
+    for event in events:
+        edges.append([event.edge[0], event.edge[1], event.time])
+    edge_index = np.array(edges)
+    srcs = np.asarray(edge_index[:, 0])
+    dsts = np.asarray(edge_index[:, 1])
+    ts = np.asarray(edge_index[:, 2])
+
+    MEMORY_MODE = 'unlimited'
+    TIME_WINDOW_RATIO = 0.15
```
Jacob-Chmura, line 22: Probably would remove this so it's clear that unlimited does not depend on time window ratio. And then check that we can really insert many events exceeding the previous capacity without issue.

test/test_model/test_edgebank.py
Jacob-Chmura, line 1: These may be lower priority, but probably worth having tests for
- ensure the right error is raised for bad args
- ensure the right default args are being set when not provided
- ensure `pos_prob` is returned correctly when customized
- ensure that unlimited memory keeps the prediction beyond the time window ratio default. I guess this is sort of tests, but we can easily add multiple events at once to test this

Jacob-Chmura (approved):
Thanks for addressing some of my comments. Still think there is a possible issue if someone tries updating the memory with empty np arrays (or np arrays whose size does not match.

A few minor tests cases could be added, but I think for a first iteration this looks good.

opendg/nn/memory/edgebank.py
shenyangHuang: agree

test/test_model/test_edgebank.py
shenyangHuang: yes, removed

test/test_model/test_edgebank.py
shenyangHuang: Thanks for these comments Jacob, these tests are quite useful to add. Added in the newest commit

opendg/nn/memory/edgebank.py
```diff
@@ -22,16 +26,19 @@ def __init__(
             pos_prob(float): the probability of the link existence for the edges in memory.
         """
         if memory_mode not in ['unlimited', 'fixed_time_window']:
-            raise ValueError(f'Invalide memory mode for EdgeBank')
+            raise ValueError(f'Invalid memory mode for EdgeBank, expected one of {...}')
```
Jacob-Chmura, line 29:
```suggestion
            raise ValueError(f'Invalid memory mode for EdgeBank, expected one of ["unlimited", "fixed"]')
```

Jacob-Chmura (approved):
Lookgs good, thanks for the updates

PR #52, opened 2025-02-27
Io/pandas

### LINKED ISSUE #47: Update IO with Pandas

Jacob-Chmura:
Close #47

### DIFF
--- a/opendg/_io/pandas.py
+++ b/opendg/_io/pandas.py
@@ -10,7 +10,7 @@ def read_pandas(
     df: pd.DataFrame,
     src_col: str,
     dst_col: str,
-    ts_col: Optional[str] = None,
+    time_col: Optional[str] = None,
     edge_feature_col: Optional[str] = None,
 ) -> List[Event]:
     events: List[Event] = []
@@ -19,8 +19,8 @@ def read_pandas(
         src_id = row[src_col]
         dst_id = row[dst_col]
 
-        if ts_col is not None:
-            time = row[ts_col]
+        if time_col is not None:
+            time = row[time_col]
         else:
             time = len(events)
 
--- a/opendg/_io/pandas.py
+++ b/opendg/_io/pandas.py
@@ -1,0 +1,38 @@
+from typing import List, Optional
+
+import pandas as pd
+import torch
+
+from opendg.events import EdgeEvent, Event
+
+
+def read_pandas(
+    df: pd.DataFrame,
+    src_col: str,
+    dst_col: str,
+    ts_col: Optional[str] = None,
+    edge_feature_col: Optional[str] = None,
+) -> List[Event]:
+    events: List[Event] = []
+
+    def _construct_event_from_row(row: pd.Series) -> None:
+        src_id = row[src_col]
+        dst_id = row[dst_col]
+
+        if ts_col is not None:
+            time = row[ts_col]
+        else:
+            time = len(events)
+
+        if edge_feature_col is not None:
+            edge_features = torch.tensor(row[edge_feature_col])
+        else:
+            edge_features = None
+
+        event = EdgeEvent(time, (src_id, dst_id), edge_features)
+        events.append(event)
+
+    # Note: This is not thread-safe. Should not be run without the GIL
+    df.apply(_construct_event_from_row, axis=1)
+
+    return events
--- a/opendg/graph.py
+++ b/opendg/graph.py
@@ -1,6 +1,8 @@
 from typing import Any, List, Optional, Union
 
-from opendg._io import read_csv, write_csv
+import pandas as pd
+
+from opendg._io import read_csv, read_pandas, write_csv
 from opendg._storage import DGStorage
 from opendg.events import Event
 from opendg.timedelta import TimeDeltaDG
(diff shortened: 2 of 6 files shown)


opendg/graph.py
```diff
@@ -45,6 +47,39 @@ def from_csv(
...
+            dst_col (str): Column in the dataframe corresponding to the destination edge id.
+            time_col (str): Column in the dataframe corresponding to the timestamp.
+            edge_feature_col (Optional[str]): Optional column in the dataframe corresponding to the edge features.
+            time_delta (Optional[TimeDeltaDG]): Describes the time granularity associated with the event stream.
+                If None, then the events are assumed 'ordered', with no specific time unit.
+
+        Returns:
+           DGraph: The newly constructed dynamic graph.
+        """
+        non_ordered_time_delta = time_delta is not None and not time_delta.is_ordered
```
Jacob-Chmura, line 74:
```suggestion
        ordered_time_delta = time_delta is None or time_delta.is_ordered
```
much simpler to read

Jacob-Chmura (commented):
Here is a quick implementation.

I think we should think carefully whether we want to include pandas IO, as it is something we will need to maintain and support, There are most likely corner cases that are not handled here, but would come up. 

I have added `pandas` as a hard package dependency for the moment. We could lazy import this (as we plan to do with `from_tgb`) though, the difference is that for pandas `DGraph` actually expects a dataframe object (instead of a string). We can still technically get away with lazy importing by skipping the type check. But at the end of the day, our library _is_ using our own pandas package in `_io/pandas.py` which makes it a hard dep.

opendg/_io/pandas.py
```diff
@@ -0,0 +1,38 @@
...
+
+from opendg.events import EdgeEvent, Event
+
+
+def read_pandas(
+    df: pd.DataFrame,
+    src_col: str,
+    dst_col: str,
+    time_col: Optional[str] = None,
+    edge_feature_col: Optional[str] = None,
```
Jacob-Chmura, line 14: These argument names make the most sense to me.

Should update them accordingly in `from_csv()`: https://github.com/shenyangHuang/openDG/blob/37fe42913ebaf067711f25e19f036e79dfc9d669/opendg/_io/csv.py#L10-L14

opendg/graph.py
```diff
@@ -45,6 +47,39 @@ def from_csv(
...
+        ordered_time_delta = time_delta is None or time_delta.is_ordered
+        if time_col is None and not ordered_time_delta:
+            raise ValueError(
+                'Must specify "time_col" when using non-ordered time delta.'
+            )
+
+        events = read_pandas(df, src_col, dst_col, time_col, edge_feature_col)
+        return cls(events, time_delta)
+
     def to_csv(self, file_path: str, *args: Any, **kwargs: Any) -> None:
```
Jacob-Chmura, line 83: Should expose the actual arguments here (as for `from_pandas()`

opendg/_io/pandas.py
```diff
@@ -0,0 +1,44 @@
+from typing import List, Optional
+
+import numpy as np
+import pandas as pd
+import torch
+
+from opendg.events import EdgeEvent, Event
+
+
+def read_pandas(
```
shenyangHuang, line 10: looks good. We should discuss the pandas dependency

opendg/graph.py
```diff
@@ -26,23 +28,65 @@ def __init__(
...
                 If None, then the events are assumed 'ordered', with no specific time unit.
-            args (Any): Optional positional arguments.
-            kwargs (Any): Optional keyword arguments.
 
         Returns:
            DGraph: The newly constructed dynamic graph.
         """
-        events = read_csv(file_path, *args, **kwargs)
+        ordered_time_delta = time_delta is None or time_delta.is_ordered
+        if time_col is None and not ordered_time_delta:
```
shenyangHuang, line 52: ordered time delta you also need to assign timestamp, it is just that sometimes you don't know what time granularity it is but you should still have timestamps.

shenyangHuang: thus I would make `time_col` mandatory

test/test_io/test_csv_io.py
```diff
@@ -71,21 +44,25 @@ def test_csv_conversion_with_features_no_feature_cols_provided():
...
     ]
+    col_names = {'src_col': 'src_id', 'dst_col': 'dst_id', 'time_col': 'time'}
 
     with tempfile.NamedTemporaryFile() as f:
         with pytest.raises(ValueError):
-            _ = write_csv(events, f.name)
+            _ = write_csv(events, f.name, **col_names)
 
 
 def test_csv_conversion_with_features_bad_feature_col_shape():
```
shenyangHuang, line 54: what does this function do?

shenyangHuang (commented):
Looks good overall, one thing is that we should make time_col mandatory (even for ordered datasets). Also, maybe adding some quick comments for helper function to remember what they do and how to use them

opendg/graph.py
Jacob-Chmura: Okay, understood. I misunderstood that ordered implies just an enumerated type (with no notion of time). If this _was_ the case, I thought the dataframe _index_ itself provides an implicit order. But I'm happy to make the required.

test/test_io/test_csv_io.py
Jacob-Chmura: This flags an error that we expect to be thrown when a user tries to read a features list with dim=37, but only 3 column names were provided (we expect 37 feature column names).

One option is to try to implicitly use _all_ other columns in the csv file as a feature dimension column. But this is fairly strict and may be annoying to deal with as a user.

Please let me know if you think other behaviour is better here.

PR #53, opened 2025-03-01
WIP: Draft dataloader

Jacob-Chmura:


opendg/loader/dataloader.py
```diff
@@ -0,0 +1,17 @@
+from opendg.graph import DGraph
+from opendg.loader.base import DGBaseLoader
+
+
+class DGDataLoader(DGBaseLoader):
+    r"""Load data from DGraph without any sampling."""
+
+    def sample(self, batch: 'DGraph') -> 'DGraph':
+        r"""DGDataLoader performs no subsampling. Reutnrs the full batch.
```
Jacob-Chmura, line 9:
```suggestion
        r"""DGDataLoader performs no subsampling. Returns the full batch.
```

opendg/timedelta.py
```diff
@@ -44,9 +44,6 @@ def __init__(self, unit: str, value: Optional[int] = 1) -> None:
         if not (isinstance(value, int)) or value <= 0:
             raise ValueError(f'Value must be a positive integer, got: {value}')
```
Jacob-Chmura, line 46: This change may not actually make sense for the purposes of storing a temporal index. It _does_ make sense in the context of defining a batch iteration type (e.g. loop through every 2 ordered events, for instance).

As the implementation get fleshed out, this may be reverted (along with the accompanying test file diff)

test/test_loader/test_dg_loader.py
Jacob-Chmura, line 1: TODO: Add unti tests

opendg/loader/base.py
Jacob-Chmura, line 1: This is pretty ugly tbh. The case of using _ordered_ batch unit on a _non-ordered_ graph is particularly prone to off-by-one errors and the sort.

Also, all the copy semantics in the `DGraph` are broken, and need to be addressed before we move forward with this implementation.

opendg/loader/base.py
```diff
@@ -0,0 +1,99 @@
...
+    def __iter__(self) -> 'DGBaseLoader':
+        return self
+
+    def __next__(self) -> 'DGraph':
+        if self._done_iteration():
+            raise StopIteration
+
+        if self._iterate_by_events:
+            events = self._events[self._idx : self._idx + self._batch_size]
+            batch = DGraph(events, time_delta=self._dg.time_delta)
```
Jacob-Chmura, line 84: Probably would be helpful to have a helper that just gives us all the edges in the graph (`dg.events`)

opendg/loader/base.py
```diff
@@ -0,0 +1,99 @@
...
+        self._idx = 0
+        self._events = []
+        if self._iterate_by_events and len(self._dg):
+            self._events = self._dg._storage.to_events()
+        elif len(self._dg):
+            assert self._dg.start_time is not None
+            self._idx = self._dg.start_time
+
+    @abstractmethod
+    def sample(self, batch: 'DGraph') -> 'DGraph':
```
Jacob-Chmura, line 65: Note: this is what we are sampling _from_ but the samples we yield will need to look into the whole dg (beyond the current batch window).

opendg/timedelta.py
```diff
@@ -122,6 +122,23 @@ class TimeDeltaUnit(str, Enum):
     def __str__(self) -> str:
         return self.value
 
+    def is_more_granular_than(self, other: 'TimeDeltaUnit') -> bool:
```
Jacob-Chmura, line 125: Without strEnum, we have to manually add a comparator since otherwise, it defaults to  string comparison (which is not what we want to compare granularity). 

I don't want to use `functools.total_ordering` since technically this enum cannot compare ordered time units

opendg/loader/base.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __next__(self) -> 'DGraph':
+        if self._done_iteration():
+            raise StopIteration
+
+        if self._iterate_by_events:
+            events = self._events[self._idx : self._idx + self._batch_size]
+            batch = DGraph(events, time_delta=self._dg.time_delta)
+        else:
+            batch = self._dg.slice_time(self._idx, self._idx + self._batch_size)
```
Jacob-Chmura, line 97: Worth noting that this creates _deep_ copies when iterating by events (which could be slow), compared to slice_time which are shallow copies.

opendg/loader/base.py
```diff
@@ -0,0 +1,125 @@
...
+        ValueError: If both the batch_unit and the graph are not TimeDeltaUnit.ORDERED, but the graph TimeDelta
+                    is coarser than the batch TimeDelta. In this case, there is ambiguity in how to yield
+                    events due to loss of information.
+
+    Note:
+        Ordered batch_unit ('TimeDeltaUnit.ORDERED) iterates using normal batch size semantics
+        in which case each yielded batch has a constant number of events (except the last batch if drop_last=False).
+            e.g. batch_size=5, batch_unit=TimeDeltaUnit.ORDERED -> yield 5 events at time
+
+        Unordered batch_unit iterates uses the appropriate temporal unit in which case each yielded
```
shenyangHuang, line 29: not related to PR but it would be great to add a tutorial showing both dataloading strategies i.e. ordered / (unordered) time-based

opendg/loader/base.py
```diff
@@ -0,0 +1,125 @@
...
+                raise ValueError(
+                    f'Tried to construct a data loader on a DGraph with time delta: {self._dg.time_delta} '
+                    f'which is strictly coarser than the batch_unit: {batch_unit}, batch_size: {batch_size}. '
+                    'Cannot iterate a non-ordered DGraph with a more granular batch_unit due to loss of informmation. '
+                    'Either choose a larger batch size, larger batch unit or consider iterate using ordered batching.'
+                )
+
+            self._batch_size = int(batch_time_delta.convert(self._dg.time_delta))
+
+        # TODO: Check for time gap?
```
shenyangHuang, line 74: Yes, in the future we can add some argument to report time gap or state it as a property, to be discussed later on

opendg/loader/base.py
shenyangHuang: pointing to self._dg

opendg/loader/base.py
shenyangHuang: not sure I understand, you mean iterating by # of events? iterating # of events is the most common use case right now, can we have pointers on dg.edges or sth? (to discuss)

test/test_loader/test_dg_loader.py
```diff
@@ -0,0 +1,259 @@
+import pytest
+
+from opendg.events import EdgeEvent
+from opendg.graph import DGraph
+from opendg.loader import DGDataLoader
+from opendg.timedelta import TimeDeltaDG, TimeDeltaUnit
+
+
```
shenyangHuang, line 8: test dataloader on `tgbl-wiki` in a function

shenyangHuang (commented):
The PR looks good overall, maybe quickly discuss it at tomorrow's meeting

opendg/loader/base.py
Jacob-Chmura: All I mean is that we create separate storage on each slice when iterating by # of events, whereas the _slice_time_ creates shallow views on the same storage as the underlying graph.

This has 2 consequences.
1. If for some reason mutating the underlying storage is supposed to be transparently reflected in this loop (e.g. some other worker appends to `DGraph`), then this _will_  _not_ be reflected when using event-based iteration.
2. If in the future some storage backend is slow to setup and teardown (e.g. some data structure with lots of pre-processing), then the iteration by events will incurr an unnecessary perf penalty.

**Note**: if the dgraph itself is _ordered_ then using an ordered batch_unit does not fall into scope of _iteration by events_ (because the time delta indexes match). **So the only time _iteration by events_ occurs, is when the graph is _not_ ordered, and the batch is ordered.**

opendg/loader/base.py
Jacob-Chmura: Yes, this is the scope of something like #40 but for data loading.

test/test_loader/test_dg_loader.py
Jacob-Chmura: I think this is more so the subject of an integration test or a perf test suite. Unit tests are meant to be more lightweight and not need any resources, so they can quickly run on GitHub CI on every pr.

In my opinion, we could have nightly / weekly builds that run integration tests, that train each type of model on a few different tgb datasets and ensures model performance is within a reasonable range.

I would also like to put together a perf test suite with sampling profilers that record the latency/throughput of each fundamental operations we support on various sized tgb datasets. This would be a great microbenchmark to use for comparing different storage engines, and can be ran manually anytime the hot path of code is touched in a pr.

PR #55, opened 2025-03-01
`DGraph` view over `DGStorage`

### LINKED ISSUE #54: Fix `DGraph` copy semantics

Jacob-Chmura:
CLose #54 #38

### DIFF
--- a/opendg/_storage/backends/array_backend.py
+++ b/opendg/_storage/backends/array_backend.py
@@ -9,8 +9,8 @@ from opendg.timedelta import TimeDeltaDG
 from ..base import DGStorageBase
 
 
-class DGStorageDictBackend(DGStorageBase):
-    r"""Dictionary implementation of temporal graph storage engine."""
+class DGStorageArrayBackend(DGStorageBase):
+    r"""Array backed implementation of temporal graph storage engine."""
 
     def __init__(self, events: List[Event]) -> None:
         self._node_feats_shape = self._check_node_feature_shapes(events)
--- a/opendg/_storage/backends/dict_backend.py
+++ b/opendg/_storage/backends/dict_backend.py
@@ -1,5 +1,4 @@
-from collections import defaultdict
-from typing import Dict, List, Optional, Set, Union
+from typing import List, Optional, Set, Union
 
 import torch
 from torch import Tensor
@@ -16,10 +15,7 @@ class DGStorageDictBackend(DGStorageBase):
     def __init__(self, events: List[Event]) -> None:
         self._node_feats_shape = self._check_node_feature_shapes(events)
         self._edge_feats_shape = self._check_edge_feature_shapes(events)
-
-        self._events_dict: Dict[int, List[Event]] = defaultdict(list)
-        for event in events:
-            self._events_dict[event.time].append(event)
+        self._events = events
 
     def to_events(
         self,
@@ -28,30 +24,25 @@ class DGStorageDictBackend(DGStorageBase):
         node_slice: Optional[Set[int]] = None,
     ) -> List[Event]:
         events: List[Event] = []
-        for t, t_events in self._events_dict.items():
-            for event in t_events:
-                if self._valid_slice(t, event, start_time, end_time, node_slice):
-                    events.append(event)
+        for event in self._events:
+            if self._valid_slice(event, start_time, end_time, node_slice):
+                events.append(event)
         return events
 
     def get_start_time(self, node_slice: Optional[Set[int]] = None) -> Optional[int]:
         start_time = None
-        for t, t_events in self._events_dict.items():
-            for event in t_events:
-                if self._valid_slice(t, event, node_slice=node_slice):
-                    if start_time is None or t < start_time:
-                        start_time = t
-                    break
+        for event in self._events:
+            if self._valid_slice(event, node_slice=node_slice):
+                if start_time is None or event.time < start_time:
+                    start_time = event.time
         return start_time
 
     def get_end_time(self, node_slice: Optional[Set[int]] = None) -> Optional[int]:
         end_time = None
-        for t, t_events in self._events_dict.items():
-            for event in t_events:
-                if self._valid_slice(t, event, node_slice=node_slice):
-                    if end_time is None or t > end_time:
-                        end_time = t
-                    break
+        for event in self._events:
+            if self._valid_slice(event, node_slice=node_slice):
+                if end_time is None or event.time > end_time:
+                    end_time = event.time
         return end_time
 
     def get_nodes(
@@ -60,15 +51,12 @@ class DGStorageDictBackend(DGStorageBase):
         end_time: Optional[int] = None,
     ) -> Set[int]:
         nodes = set()
-        for t, t_events in self._events_dict.items():
-            for event in t_events:
-                if self._valid_slice(
-                    t, event, start_time=start_time, end_time=end_time
-                ):
-                    if isinstance(event, NodeEvent):
-                        nodes.add(event.node_id)
-                    elif isinstance(event, EdgeEvent):
-                        nodes.union(event.edge)
+        for event in self._events:
+            if self._valid_slice(event, start_time=start_time, end_time=end_time):
+                if isinstance(event, NodeEvent):
+                    nodes.add(event.node_id)
+                elif isinstance(event, EdgeEvent):
+                    nodes.union(event.edge)
         return nodes
 
     def get_num_edges(
@@ -78,16 +66,14 @@ class DGStorageDictBackend(DGStorageBase):
         node_slice: Optional[Set[int]] = None,
     ) -> int:
         edges = set()
-        for t, t_events in self._events_dict.items():
-            for event in t_events:
-                if isinstance(event, EdgeEvent) and self._valid_slice(
-                    t,
-                    event,
-                    start_time=start_time,
-                    end_time=end_time,
-                    node_slice=node_slice,
-                ):
-                    edges.add((event.time, event.edge))
+        for event in self._events:
+            if isinstance(event, EdgeEvent) and self._valid_slice(
+                event,
+                start_time=start_time,
+                end_time=end_time,
+                node_slice=node_slice,
+            ):
+                edges.add((event.time, event.edge))
         return len(edges)
 
     def get_num_timestamps(
@@ -97,17 +83,14 @@ class DGStorageDictBackend(DGStorageBase):
         node_slice: Optional[Set[int]] = None,
     ) -> int:
         timestamps = set()
-        for t, t_events in self._events_dict.items():
-            for event in t_events:
-                if self._valid_slice(
-                    t,
-                    event,
-                    start_time=start_time,
-                    end_time=end_time,
-                    node_slice=node_slice,
-                ):
-                    timestamps.add(t)
-                    break
+        for event in self._events:
+            if self._valid_slice(
+                event,
+                start_time=start_time,
+                end_time=end_time,
+                node_slice=node_slice,
+            ):
+                timestamps.add(event.time)
         return len(timestamps)
 
     def append(self, events: Union[Event, List[Event]]) -> 'DGStorageBase':
@@ -115,7 +98,7 @@ class DGStorageDictBackend(DGStorageBase):
             events = [events]
 
         # Check that the new events have matching feature dimension
-        if len(self._events_dict):
+        if len(self._events):
             # Node/edge feature shape must match our current feature shape
             exp_node_feats_shape = self._node_feats_shape
             exp_edge_feats_shape = self._edge_feats_shape
@@ -134,8 +117,7 @@ class DGStorageDictBackend(DGStorageBase):
             events, expected_shape=exp_edge_feats_shape
         )
 
-        for event in events:
-            self._events_dict[event.time].append(event)
+        self._events += events
         return self
 
     def temporal_coarsening(
@@ -153,20 +135,18 @@ class DGStorageDictBackend(DGStorageBase):
         max_time, max_node_id = -1, -1
 
         indices, values = [], []
-        for t, t_events in self._events_dict.items():
-            for event in t_events:
-                if isinstance(event, NodeEvent) and self._valid_slice(
-                    t,
-                    event,
-                    start_time=start_time,
-                    end_time=end_time,
-                    node_slice=node_slice,
-                ):
-                    indices.append([event.time, event.node_id])
-                    values.append(event.features)
+        for event in self._events:
+            if isinstance(event, NodeEvent) and self._valid_slice(
+                event,
+                start_time=start_time,
+                end_time=end_time,
+                node_slice=node_slice,
+            ):
+                indices.append([event.time, event.node_id])
+                values.append(event.features)
 
-                    max_time = max(max_time, t)
-                    max_node_id = max(max_node_id, event.node_id)
+                max_time = max(max_time, event.time)
+                max_node_id = max(max_node_id, event.node_id)
 
         if not len(values):
             return None
@@ -191,20 +171,18 @@ class DGStorageDictBackend(DGStorageBase):
         max_time, max_node_id = -1, -1
 
         indices, values = [], []
-        for t, t_events in self._events_dict.items():
-            for event in t_events:
-                if isinstance(event, EdgeEvent) and self._valid_slice(
-                    t,
-                    event,
-                    start_time=start_time,
-                    end_time=end_time,
-                    node_slice=node_slice,
-                ):
-                    indices.append([event.time, event.edge[0], event.edge[1]])
-                    values.append(event.features)
+        for event in self._events:
+            if isinstance(event, EdgeEvent) and self._valid_slice(
+                event,
+                start_time=start_time,
+                end_time=end_time,
+                node_slice=node_slice,
+            ):
+                indices.append([event.time, event.edge[0], event.edge[1]])
+                values.append(event.features)
 
-                    max_time = max(max_time, t)
-                    max_node_id = max(max_node_id, event.edge[0], event.edge[1])
+                max_time = max(max_time, event.time)
+                max_node_id = max(max_node_id, event.edge[0], event.edge[1])
 
         if not len(values):
             return None
@@ -227,7 +205,6 @@ class DGStorageDictBackend(DGStorageBase):
 
     def _valid_slice(
         self,
-        time: int,
         event: Event,
         start_time: Optional[int] = None,
         end_time: Optional[int] = None,
@@ -236,7 +213,7 @@ class DGStorageDictBackend(DGStorageBase):
         lb_time = float('-inf') if start_time is None else start_time
         ub_time = float('inf') if end_time is None else end_time
 
-        time_valid = lb_time <= time < ub_time
+        time_valid = lb_time <= event.time < ub_time
         node_valid = (
             node_slice is None
             or (isinstance(event, NodeEvent) and event.node_id in node_slice)
--- a/opendg/_storage/backends/dict_backend.py
+++ b/opendg/_storage/backends/dict_backend.py
@@ -138,24 +138,31 @@ class DGStorageDictBackend(DGStorageBase):
 
         indices, values = [], []
         for event in self._events:
-            if (
-                isinstance(event, NodeEvent)
-                and event.features is not None
-                and self._valid_slice(
-                    event,
-                    start_time=start_time,
-                    end_time=end_time,
-                    node_slice=node_slice,
-                )
+            if self._valid_slice(
+                event, start_time=start_time, end_time=end_time, node_slice=node_slice
             ):
-                indices.append([event.time, event.node_id])
-                values.append(event.features)
+                if isinstance(event, NodeEvent):
+                    max_time = max(max_time, event.time)
+                    max_node_id = max(max_node_id, event.node_id)
 
-                max_time = max(max_time, event.time)
-                max_node_id = max(max_node_id, event.node_id)
+                    if event.features is not None:
+                        indices.append([event.time, event.node_id])
+                        values.append(event.features)
+
+                elif isinstance(event, EdgeEvent):
+                    max_time = max(max_time, event.time)
+                    max_node_id = max(max_node_id, event.edge[0], event.edge[1])
 
         if not len(values):
             return None
+
+        # If the end_time is given, then it determines the dimension of the temporal axis
+        # This is true even if there are no events at the end time, which could occur after
+        # calling slice_time on a graph.
+        if end_time is not None:
+            max_time = end_time
+        else:
+            max_time = max_time + 1
 
         values_tensor = torch.stack(values)
         indices_tensor = torch.tensor(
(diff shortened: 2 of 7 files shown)


opendg/graph.py
```diff
@@ -56,22 +56,30 @@ def to_csv(self, file_path: str, *args: Any, **kwargs: Any) -> None:
...
         Args:
             start_time (int): The start of the temporal slice.
             end_time (int): The end of the temporal slice (exclusive).
+
+        Returns:
+            DGraph copy of events between start and end_time.
         """
-        self._storage.slice_time(start_time, end_time)
+        events = self._storage.slice_time(start_time, end_time).to_events()
+        return DGraph(events, time_delta=self.time_delta)
```
Jacob-Chmura, line 70: It's a shame we cannot construct `DGraph` directly from storage, as this `.to_events()` basically forces us into making an extra copy, and iterating over the events list an extra time.

We could of course add `DGStorageBase` as an optional argument into the constructor, though I don't like the side-effect of polluting the public API. Alternatively, we could mess around with `cls.__new__()` or something similar, but this seems overly complicated. Leaving this as is for now.

Jacob-Chmura:
This is needed for #28

opendg/_storage/backends/dict_backend.py
```diff
@@ -13,57 +13,109 @@
...
-        self._num_nodes: Optional[int] = None
-        self._num_edges: Optional[int] = None
-        self._num_timestamps: Optional[int] = None
-
-    def to_events(self) -> List[Event]:
+    def to_events(
+        self,
+        start_time: Optional[int] = None,
+        end_time: Optional[int] = None,
+        node_slice: Optional[Set[int]] = None,
```
Jacob-Chmura, line 28: Perhaps this is worth encapsulating in an object to make it more clear. But the idea is that the `DGraph` maintains `start_time`, `end_time`, `node_slice` across shallow copies of slicing ops, and then forwards them to the storage as _constraints_. The storage is responsible for answering the query, but does not make any modification to the underlying store (by default).

Start/end time are upper and lower bounds on the time constraints (as a result of `.slice_time()` and node slice is a set of nodes to constrain on (as a result of `.slice_nodes()`)

opendg/_storage/backends/dict_backend.py
```diff
@@ -13,57 +13,109 @@
...
+    ) -> List[Event]:
         events: List[Event] = []
-        for t_events in self._events_dict.values():
-            events += t_events
+        for t, t_events in self._events_dict.items():
+            for event in t_events:
+                if self._valid_slice(t, event, start_time, end_time, node_slice):
```
Jacob-Chmura, line 33: As I mention in a comment, this can be sped up by early exiting all events on the current time stamp if the time stamp is not valid. This can be achieved by having separate functions for `valid_time_slice` and `valid_node_slice` or just returning a tuple out.

However, it should be clear from the code, that the dictionary backend is effectively `O(number of events)` for every query that is a cache miss at the `DGraph` level.

opendg/_storage/base.py
```diff
@@ -1,121 +1,88 @@
...
-    def num_edges(self) -> int:
-        r"""The total number of unique edges encountered over the dynamic graph."""
+    def append(self, events: Union[Event, List[Event]]) -> 'DGStorageBase':
+        pass
 
-    @property
     @abstractmethod
-    def num_timestamps(self) -> int:
-        r"""The total number of unique timestamps encountered over the dynamic graph."""
+    def temporal_coarsening(
```
Jacob-Chmura, line 64: Since the time delta is now stored at the `DGraph` level, we will need to update this to take two arguments (the original time delta, and the one we are converting to).

opendg/graph.py
```diff
@@ -53,25 +73,59 @@ def to_csv(self, file_path: str, *args: Any, **kwargs: Any) -> None:
...
             start_time (int): The start of the temporal slice.
             end_time (int): The end of the temporal slice (exclusive).
+
+        Returns:
+            DGraph view of events between start and end_time.
         """
-        self._storage.slice_time(start_time, end_time)
+        self._check_slice_time_args(start_time, end_time)
+
+        dg = copy.copy(self)
```
Jacob-Chmura, line 95: These are shallow copies. User needs to be aware of this to avoid side-effects on the underlying storage.

test/test_storage_impl.py
Jacob-Chmura, line 1: TODO: Implement these tests and then update tests for `DGraph`

test/test_dgraph.py
```diff
@@ -13,3 +13,64 @@ def test_init_empty_dgraph_with_time_delta():
     td = TimeDeltaDG(unit='ms', value=5)
     dgraph = DGraph(events, td)
     assert dgraph.time_delta == td
+
+
+"""
```
Jacob-Chmura, line 18: These are the set of test cases that need to be implemented on the `DGraph`

opendg/graph.py
```diff
@@ -91,42 +138,143 @@ def temporal_coarsening(
...
+        Returns a tensor.sparse_coo_tensor of size T x V x V x d where
+
+        - T = Number of timestamps
+        - E = Number of edges
+        - d = Edge feature dimension
+
+        or None if there are no edge features on the dynamic graph.
+        """
+        if self._cache.get('edge_feats') is None:
+            self._cache['edge_feats_feats'] = self._storage.get_edge_feats(
```
Jacob-Chmura, line 240:
```suggestion
            self._cache['edge_feats'] = self._storage.get_edge_feats(
```

shenyangHuang (commented):
added some comments, most important one is that if the DGraph is sliced by time or by node, shouldn't `num_nodes`, `num_edges`, `num_timestamps` be updated as well.

opendg/_storage/backends/dict_backend.py
```diff
@@ -13,57 +12,96 @@
 class DGStorageDictBackend(DGStorageBase):
     r"""Dictionary implementation of temporal graph storage engine."""
 
-    def __init__(self, events: List[Event], time_delta: TimeDeltaDG) -> None:
+    def __init__(self, events: List[Event]) -> None:
```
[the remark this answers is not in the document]
Jacob-Chmura: Correct. As you can see, it's not actually used in any storage ops anymore. The tome delta tells us how to _interpret_ the data in the storage, so I think it makes more sense to hold at the Dgraph level.

opendg/_storage/backends/dict_backend.py
```diff
@@ -13,57 +12,96 @@
...
-                elif isinstance(event, EdgeEvent) and len(
-                    set(event.edge).intersection(nodes)
-                ):
-                    events_dict[t].append(event)
-        self._events_dict = events_dict
-        self._invalidate_cache()
-        return self
-
-    def append(self, events: Union[Event, List[Event]]) -> 'DGStorageBase':
+    def get_start_time(self, node_slice: Optional[Set[int]] = None) -> Optional[int]:
```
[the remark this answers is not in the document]
Jacob-Chmura: Correct. The first time we load the graph is a cache miss, and we compute the start time.

From then on, we read the cached result. After slicing time, we update the cache value (without recomputing).

Upon slicing nodes, we invalidate the cache (because the new sliced graph may have a later start time).

opendg/_storage/backends/dict_backend.py
```diff
@@ -13,57 +12,96 @@
...
+                    nodes.add(event.edge[1])
+        return nodes
+
+    def get_num_edges(
+        self,
+        start_time: Optional[int] = None,
+        end_time: Optional[int] = None,
+        node_slice: Optional[Set[int]] = None,
+    ) -> int:
+        edges = set()
```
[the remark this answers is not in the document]
Jacob-Chmura: Sounds good, yes this is an implementation choice to be discussed. I have kept the behaviour that was previously implemented.

opendg/_storage/backends/dict_backend.py
```diff
@@ -82,119 +120,118 @@ def append(self, events: Union[Event, List[Event]]) -> 'DGStorageBase':
...
-
-    @property
-    def num_timestamps(self) -> int:
-        if self._num_timestamps is None:
-            self._num_timestamps = len(self._events_dict)
-        return self._num_timestamps
-
-    @property
-    def node_feats(self) -> Optional[Tensor]:
+    def get_node_feats(
```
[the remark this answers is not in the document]
Jacob-Chmura: I have (kind of intentionally) dropped documentation from our private storage implementations, since it's largely duplicated and shadowed by the public Dgraoh api.

The private classes are not being flagged on read the docs either.

opendg/_storage/backends/dict_backend.py
```diff
@@ -82,119 +120,118 @@ def append(self, events: Union[Event, List[Event]]) -> 'DGStorageBase':
...
             indices
         ).t()  # https://pytorch.org/docs/stable/sparse.html#construction
 
+        assert self._edge_feats_shape is not None
+
         shape = (
-            self.end_time + 1,
-            self.num_nodes,
-            self.num_nodes,
+            max_time + 1,
```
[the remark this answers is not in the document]
Jacob-Chmura: Yes we do. You can check the tests for more clear examples, but the high level behaviour id that we always use the max node Id in the current sliced graph. But this node is still local to the storage frame.

There is probably some global/local id mapping that needs to take place here. Let's discuss tomorrow.

opendg/graph.py
```diff
@@ -91,42 +138,143 @@ def temporal_coarsening(
...
+    def time_delta(self) -> TimeDeltaDG:
         r"""The time granularity of the dynamic graph."""
-        return self._storage.time_delta
+        return self._time_delta
 
     @property
     def num_nodes(self) -> int:
         r"""The total number of unique nodes encountered over the dynamic graph."""
-        return self._storage.num_nodes
+        if self._cache.get('num_nodes') is None:
```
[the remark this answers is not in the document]
Jacob-Chmura: Yes it does. I invalidate this cache value when slicing.

opendg/graph.py
```diff
@@ -53,25 +64,59 @@ def to_csv(self, file_path: str, *args: Any, **kwargs: Any) -> None:
...
+            DGraph copy of events related to the input nodes.
         """
-        self._storage.slice_nodes(nodes)
+        dg = copy.copy(self)
+        dg._cache.clear()
+
+        if self._cache.get('node_slice') is None:
+            self._cache['node_slice'] = set(range(self.num_nodes))
+
+        dg._cache['node_slice'] = self._cache['node_slice'] & set(nodes)
```
Jacob-Chmura, line 117: @shenyangHuang See here, we force cache refresh when slicing nodes.

This means all the properties become None and we are forced to re-compute from storage. 

The new Dgraph copy gets node slice set to the intersection of the current sliced graph and the query nodes. (In case the current graph cache is empty, we take _all_ the nodes of the current graph to slice)

Technically, this could further be optimized (maybe the slice is the full graph). But this is a corner case that I assume never happens. For all intents and purposes, whenever we slice nodes, all our stats need a refresh, which is handled here.

Jacob-Chmura:
> added some comments, most important one is that if the DGraph is sliced by time or by node, shouldn't `num_nodes`, `num_edges`, `num_timestamps` be updated as well. 


Thanks, Andy! See my inline replies addressing this.

opendg/graph.py
```diff
@@ -78,33 +88,115 @@ def from_pandas(
...
+
+        start_time_with_node_slice = self._storage.get_start_time(
+            dg._cache.get('node_slice')
+        )
+        if self._cache.get('start_time') is None:
```
Jacob-Chmura, line 175: These cache updates are quite difficult to track and get right. Now that the testing is in place, I can look through this and try to simplify.

test/test_dgraph.py
Jacob-Chmura, line 1: I will review this and see if certain test helpers can simplify the test suite, but I am doubtful that any properly exhaustive test of all these operations can be done in anything under close to 1k lines

Jacob-Chmura (commented):
There are still plenty of edge cases that came up while building out the test suite. Many of these I have chosen the simplest behaviour, but we should make sure to review this otherwise the client may have a less than intuitive time trying to understanding our `DGraph`.

A few things off the top of my head that I recall threw my off guard a bit when writing the tests
- Generally, the fact that `len(DGraph)` has nothing to do with `start_time` and `end_time` and everything to do with the number of events in the graph is sometime confusing. For instance, when we have `start_time < end_time` but `len() == 0` or `start_time == end_time` but `len() > 0`.
- The fact that empty graphs implies `start_time is None` and `end_time is None` but empty in this context means _no events_. In particular, the converse is not true if we slice away a graph so that no events remain, length is 0, but we still have a left over start/end time based on our slice
- The fact that slicing does not _clip_ to the boundary of a graph. For instance, if we slice a graph with an event at time 1 and time 10 using `slice_time(1, 5)` we end up with `end_time = 4` (because we had made our slicing end range exclusive as with most container operators.
- The fact that nodes can _come back to life_. Currently, we consider all the nodes of a particular edge to be alive (part of the graph). So say we have a single `EdgeEvent(src=2, dst=4)`. If we call `slice_nodes({4})`, we will find that node 2 is still a part of the active node slice, because it belong to an edge with `4`, which is part of the slice. This is particularly confusing when we explicitly slice away node 2 in some previous step, and then see it come back to the graph.
- The above is also made slightly more confusing by the fact that we, at all times, consider the number of nodes in the graph to be 1 + max_node_id (the 1 + is for 0-indexing). So in the previous example, with node id 4, we claim to have 5 nodes in the graph (0, 1, 2, 3, 4). The reason I saw this is confusing, is that a slice operation can appear to _add_ nodes to the graph! This happens if the _phantom_ node problem occurs (where we sliced away a node and then it came back since it was an edge with a different node within the slice), specifically to a node who's `id` is the largest in the graph.

These are just some of my observations. Will work on trying to simplify the test cases (insofar as that is possible), I implore reviewer to check out the test cases and see if something does not make sense. We can change the behaviour, but in the current state, fixing the test cases is quite tedious.

opendg/_storage/backends/array_backend.py
```diff
@@ -0,0 +1,253 @@
...
+        node_valid = (
+            node_slice is None
+            or (isinstance(event, NodeEvent) and event.node_id in node_slice)
+            or (
+                isinstance(event, EdgeEvent)
+                and len(set(event.edge).intersection(node_slice)) > 0
+            )
+        )
+        # TODO: This can be optimized by returning these seperately, and hence early
+        # returning out of the event loop if we already know the timestamp is not valid
```
shenyangHuang, line 252: yes agree, I think `time_valid` can be checked much quicker than `node_valid` especially if the event has many NodeEvent / EdgeEvent (i.e. for a very dense day). We should return `time_valid` false early.

opendg/_storage/backends/array_backend.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def to_events(
+        self,
+        start_time: Optional[int] = None,
+        end_time: Optional[int] = None,
+        node_slice: Optional[Set[int]] = None,
+    ) -> List[Event]:
+        events: List[Event] = []
+        for event in self._events:
+            if self._valid_slice(event, start_time, end_time, node_slice):
```
shenyangHuang, line 30: This is much simplified, definitely better now

opendg/_storage/backends/array_backend.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def get_end_time(self, node_slice: Optional[Set[int]] = None) -> Optional[int]:
+        end_time = None
+        for event in self._events:
+            if self._valid_slice(event, node_slice=node_slice):
+                if end_time is None or event.time > end_time:
+                    end_time = event.time
+        return end_time
+
+    def get_nodes(
```
shenyangHuang, line 50: to discuss: do we need a `get_edges()` function similar to this one.

opendg/_storage/base.py
```diff
@@ -12,110 +12,80 @@ class DGStorageBase(ABC):
...
+        start_time: Optional[int] = None,
+        end_time: Optional[int] = None,
+        node_slice: Optional[Set[int]] = None,
+    ) -> List[Event]:
+        pass
 
     @abstractmethod
-    def slice_time(self, start_time: int, end_time: int) -> 'DGStorageBase':
-        r"""Extract temporal slice of the dynamic graph between start_time and end_time."""
+    def get_start_time(self, node_slice: Optional[Set[int]] = None) -> Optional[int]:
```
shenyangHuang, line 28: if there is no start time, should this return `None` or `-inf`

opendg/_storage/base.py
```diff
@@ -12,110 +12,80 @@ class DGStorageBase(ABC):
...
+        pass
 
     @abstractmethod
-    def slice_nodes(self, nodes: List[int]) -> 'DGStorageBase':
-        r"""Extract topological slice of the dynamcic graph given the list of nodes."""
+    def get_end_time(self, node_slice: Optional[Set[int]] = None) -> Optional[int]:
```
shenyangHuang, line 32: same here, `None` or `inf`

opendg/graph.py
```diff
@@ -124,42 +225,155 @@ def temporal_coarsening(
...
+        """
+        if self._cache.get('edge_feats') is None:
+            self._cache['edge_feats'] = self._storage.get_edge_feats(
+                self._cache.get('start_time'),
+                self._cache.get('end_time'),
+                self._cache.get('node_slice'),
+            )
+        return self._cache['edge_feats']
+
+    def _check_storage_args(
```
shenyangHuang, line 348: if this is only used in `__init__` should we just keep it in `__int__` ?

opendg/graph.py
```diff
@@ -124,42 +225,155 @@ def temporal_coarsening(
...
+        self,
+        events: Optional[List[Event]] = None,
+        _storage: Optional[DGStorageBase] = None,
+    ) -> 'DGStorageBase':
+        if _storage is not None:
+            if events is not None:
```
shenyangHuang, line 354: User can't specify `DGStorage` because the backend type is not accessible? I don't quite get this error.

opendg/graph.py
```diff
@@ -18,11 +20,19 @@ class DGraph:
...
+        _storage: Optional[DGStorageBase] = None,
     ) -> None:
+        self._storage = self._check_storage_args(events, _storage)
+
         if time_delta is None:
-            time_delta = TimeDeltaDG('r')  # Default to ordered if granularity
-        self._storage = DGStorage(events, time_delta)
+            self._time_delta = TimeDeltaDG('r')  # Default to ordered granularity
+        else:
+            self._time_delta = time_delta
```
shenyangHuang, line 33: we should check the type of `time_delta` here and enforce it to be `TimeDeltaDG` otherwise we will run into issue.

opendg/graph.py
```diff
@@ -78,33 +88,114 @@ def from_pandas(
...
+            DGraph view of events between start and end_time.
         """
-        self._storage.slice_time(start_time, end_time)
+        self._check_slice_time_args(start_time, end_time)
+
+        dg = copy.copy(self)
+        dg._cache = copy.deepcopy(self._cache)  # Deep copy cache to avoid dict alias
+
+        new_start_time = start_time
+        new_end_time = end_time - 1  # Because slicing is end range exclusive
```
shenyangHuang, line 131: so basically when a user do sth like 
```
st = dg.start_time
et = dg.end_time
new_dg = dg.slice_time(st, et) 
```
the `new_dg` becomes shorter in duration??

opendg/graph.py
```diff
@@ -78,33 +88,114 @@ def from_pandas(
...
+            new_end_time = self.end_time
 
-    def slice_nodes(self, nodes: List[int]) -> None:
+        # Force cache refresh on the new copy if we actually sliced the graph
+        if new_start_time != self.start_time or new_end_time != self.end_time:
+            dg._cache.clear()
+            dg._cache['node_slice'] = self._cache.get('node_slice')
+
+        dg._cache['start_time'] = new_start_time
+        dg._cache['end_time'] = new_end_time + 1  # End-range exclusive value in cache
```
shenyangHuang, line 145: the `+1` and `-1` is confusing to me

test/test_dgraph.py
shenyangHuang: sounds good

shenyangHuang (commented):
added some comments / suggestions

opendg/_storage/backends/array_backend.py
Jacob-Chmura: Agreed. Is this blocking? Perhaps it can be addressed in a future PR when we have a compelling example of needing it. This PR is already quite big. If we know we definitely need it then it can be added here I suppose

opendg/_storage/base.py
Jacob-Chmura: Yes, technically user can construct an event with such a timestamp, in which case we would have to check `len()` to distinguish empty graph. Other than that I think using `-inf` and `inf` sentinel markers to denote time stamps on empty graphs, but there may be some other corner cases that arise. It sure does simplify some of the implementation to not have to check for `None`.

opendg/graph.py
Jacob-Chmura: The idea is that a user cannot simultaneously initialize a `DGraph` using the copy constructor, and using new events. Either `_storage` is specified (and `events` is empty), in which case we are creating a new view over an existing graph, or `events` is given but `_storage` is empty, in which case we are creating a new `DGStorage` backed graph.

If both are given, this is undefined (technically we could define the behaviour to mean _append to the existing storage_). But the user is never supposed to manipulate `DGStorage` so this is what we want to enforce during `DGraph` construction.

opendg/graph.py
Jacob-Chmura: Done

opendg/graph.py
Jacob-Chmura: Correct. This is not the most intuitive. We could change the slice to be end-range inclusive if you think this is best. The original implementation was following `__slice__` semantics, which always return half-open intervals (e.g. for lists, numpy arrays or tensors)

opendg/graph.py
Jacob-Chmura: Yeah. The idea is that our `end_time` cache value is the _inclusive_ end time. This is the end time that corresponds to the max timestamp on the graph. On the other hand, all slice ops (including the ones passed forward to the `DGStorage` for checking `valid_slice` are _exclusive_ of the end time. 

If we decide to always use inclusive end time semantics this would simplify a bit.

opendg/graph.py
```diff
@@ -78,33 +88,114 @@ def from_pandas(
...
+
+        return dg
+
+    def slice_nodes(self, nodes: List[int]) -> 'DGraph':
         r"""Extract topological slice of the dynamcic graph given the list of nodes.
```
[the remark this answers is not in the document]
Jacob-Chmura: Sounds good. And analogous should is relevant when slicing nodes (the start/end time are implicitly updated to the new min/max end time on the graph. 

But again, the latter is an implementation detail which is not clear to me is the right one. It would take a bit of time to refactor the test cases in case we switch this but its worth getting right. Do you think the start/end times of the graph should be updated when slicing nodes (as in the current implementation)? Or should they be effectively, read-only?

PR #57, opened 2025-03-10
Add `TimeDeltaUnit` enum

### LINKED ISSUE #56: Add Enumeration for `TimeDeltaDG` units

Jacob-Chmura:
Close #56 

### Main Purpose
The purpose of this PR is to add a `TimeDeltaUnit` enumeration that holds all valid temporal unit values in a string enum. This provides two major benefits:
- simplifies checking if a temporal unit is one of the valid units we support, or ordered (e.g. `unit == TimeDelta.ORDERED`)
- provides a native ordering of temporal units which will be handy for certain conversions (e.g. `TimeDelta.SECOND < TimeDelta.WEEK`)

### Note
The [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) native to python was introduced in _3.11_ (we are on _3.9_), hence we emulate it using multiple inheritance.

### DIFF
--- a/opendg/timedelta.py
+++ b/opendg/timedelta.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from typing import Optional, Union
 
 
@@ -14,6 +15,7 @@ class TimeDeltaDG:
             'r':  Ordered granularity (no conversions allowed, TimeDelta value must be 1)
             'Y':  Year
             'M':  Month
+            'W':  Week
             'D':  Day
             'h':  Hour
             'm':  Minute
@@ -24,7 +26,6 @@ class TimeDeltaDG:
     """
 
     def __init__(self, unit: str, value: Optional[int] = 1) -> None:
-        self._ordered_unit = 'r'
         self._unit_nano_ratio = {
             'Y': 1000 * 1000 * 1000 * 60 * 60 * 24 * 365,
             'M': 1000 * 1000 * 1000 * 60 * 60 * 24 * 30,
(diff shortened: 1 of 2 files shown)


opendg/timedelta.py
```diff
@@ -104,3 +102,35 @@ def _convert_from_delta(self, other: 'TimeDeltaDG') -> float:
...
+    r"""Temporal unit of time granularity."""
+
+    ORDERED = 'r'
+    NANOSECOND = 'ns'
+    MICROSECOND = 'us'
+    MILLISECOND = 'ms'
+    SECOND = 's'
+    MINUTE = 'm'
+    HOUR = 'h'
+    DAY = 'D'
```
shenyangHuang, line 117: why are day, week, etc. caps?

test/test_timedelta.py
```diff
@@ -34,6 +34,24 @@ def test_init_ordered():
     assert td.is_ordered
 
 
+def test_init_with_time_delta_unit(temporal_unit):
+    temporal_unit = TimeDeltaUnit(temporal_unit)
```
shenyangHuang, line 38: `temporal_unit` maybe should be called `time_granularity`

shenyangHuang (approved):
PR looks good to me

opendg/timedelta.py
Jacob-Chmura: This is sort of standard convention that I've seen for enums, to communicate that these are constants. Though this is mostly private and the user is mostly meant to instantiate from the value strings ('r', 'ns', etc)

test/test_timedelta.py
Jacob-Chmura: sure

PR #64, opened 2025-03-12
WIP: Add get_nbrs to dgraph

Jacob-Chmura:


opendg/_storage/backends/array_backend.py
```diff
@@ -130,6 +131,43 @@ def temporal_coarsening(
     ) -> 'DGStorageBase':
         raise NotImplementedError('Temporal Coarsening is not implemented')
 
+    def get_nbrs(
+        self,
+        seed_nodes: List[int],
+        num_nbrs: List[int],
+        start_time: Optional[int] = None,
+        end_time: Optional[int] = None,
+        node_slice: Optional[Set[int]] = None,
```
Jacob-Chmura, line 140: If we want to support non-uniform sampling, this will have to take a sampling function as an argument.

opendg/_storage/backends/array_backend.py
```diff
@@ -130,6 +131,43 @@ def temporal_coarsening(
...
+    ) -> Dict[int, List[List[int]]]:
```
Jacob-Chmura, line 141: Is this the right format for our return type? This gives list of lists as return, but if we only need the aggregated hops (and not the intermediate ones), then this could be simplified to `Dict[int, List[int]]`

opendg/_storage/base.py
```diff
@@ -69,6 +69,17 @@ def temporal_coarsening(
     ) -> 'DGStorageBase':
         pass
 
+    @abstractmethod
```
Jacob-Chmura, line 72: I have a strong feeling that the nbr logic should be kept in the storage backend for non-online samplers (i.e. the ones that aren't simple variants of the `DGraph` iterator.

We need at least 1-hop info here. But then in my view we might aswell just put multi-hop logic here as well, since it's possible for the storage to optimize for multi-hops (because it is aware of the internal representation), whereas putting it outside the storage, prevents such optimizations).

opendg/graph.py
```diff
@@ -240,6 +240,33 @@ def temporal_coarsening(
         """
         raise NotImplementedError
 
+    def get_nbrs(
```
Jacob-Chmura, line 243: This means we directly can get neighbours on the `DGraph` view itself. How should this interact with different samplers?

Presumably we don't need to change the sampling technique at runtime? But if it's all handled at the storage level then we need this information at construction time (when the storage is created).

opendg/_storage/base.py
```diff
@@ -69,6 +69,17 @@ def temporal_coarsening(
...
+    def get_nbrs(
+        self,
+        seed_nodes: List[int],
+        num_nbrs: List[int],
```
Jacob-Chmura, line 76: What if we don't want to do any sampling? Put `num_nbrs[i] = -1` or `None`?

shenyangHuang (commented):
Look good to me. The only point to finalize is the return type and I agree that neighborsampling can be done at Storage

PR #68, opened 2025-03-13
Fix slow dgraph copies

### LINKED ISSUE #67: Avoid using __copy__ on  `DGraph`

Jacob-Chmura:
Close #67 

### Purpose
This PR removes all `__copy__` calls in favour of our fast _copy constructor_ when performing shallow clones of the `DGraph`.

### Perf Results
Iterating a `DGraph` with 1M events:

#### Control (`openDG::main`)
control

- notice we spend about 16s out of the 21s in `slice_time` making these copies.

#### Case (`openDG::bug/slow_copy`)
case

- not anymore

### DIFF
--- a/opendg/graph.py
+++ b/opendg/graph.py
@@ -1,4 +1,3 @@
-import copy
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


shenyangHuang (approved):
looks great, and nice performance improvement. Thanks Jacob.

PR #72, opened 2025-03-18
Bump python 3.10

Jacob-Chmura:


shenyangHuang:
No issue from my end, I might need to update my local version then

PR #80, opened 2025-03-18
Simplify io interface

### LINKED ISSUE #79: Unified IO DGraph Constructor

Jacob-Chmura:
Close #79

### DIFF
--- a/opendg/_io/__init__.py
+++ b/opendg/_io/__init__.py
@@ -1,2 +1,18 @@
-from opendg._io.csv import read_csv, write_csv
+from opendg._io.csv import read_csv
 from opendg._io.pandas import read_pandas
+
+from typing import Any, List
+import pathlib
+
+import pandas as pd
+
+from opendg.events import Event
+
+
+def read_events(data: str | pathlib.Path | pd.DataFrame, **kwargs: Any) -> List[Event]:
+    if isinstance(data, pd.DataFrame):
+        return read_pandas(data, **kwargs)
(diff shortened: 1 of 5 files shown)


opendg/graph.py
```diff
@@ -1,106 +1,35 @@
...
-    """
+    r"""Dynamic Graph Object provides a 'view' over a DGStorage backend."""
 
     def __init__(
         self,
-        events: Optional[List[Event]] = None,
-        time_delta: Optional[TimeDeltaDG] = None,
-        _storage: Optional[DGStorageBase] = None,
+        data: DGStorage | List[Event] | str | pathlib.Path | pd.DataFrame,
+        time_delta: TimeDeltaDG = TimeDeltaDG(unit='r'),
```
Jacob-Chmura, line 19: This is ok because `TimeDeltaDG` if `__frozen__` and thus immutable

opendg/_io/__init__.py
```diff
@@ -1,2 +1,18 @@
...
+import pathlib
+
+import pandas as pd
+
+from opendg.events import Event
+
+
+def read_events(data: str | pathlib.Path | pd.DataFrame, **kwargs: Any) -> List[Event]:
+    if isinstance(data, pd.DataFrame):
+        return read_pandas(data, **kwargs)
```
Jacob-Chmura, line 14: Will need to do something like:
```python
elif isinstance(data, str) and data.isin(TGB_DATASETS) 
```
or check for csv file ending so that we know if the input is meant to be read from disc or via tgb

PR #89, opened 2025-03-19
Take advantage of sorted events in `ArrayBackend`

### LINKED ISSUE #88: Take Advantage of Sorted Events in `ArrayBackend`

Jacob-Chmura:
Close #88

### Purpose
The purpose of this PR is to use the sortedness of the events list to speed up and simplify the `ArrayBackend`. I also made several micro optimizations which you can read in the code.

### Complexity
The complexity is clearly better but we would need to add more exhaustive perf testing to truly see the benefits.

All time-dependant storage queries are reduced from $$\mathcal{O}(|Events|)$$ down to $$\mathcal{O}(|time range|)$$ (plus an additional overhead of $$\mathcal{O}|log|T||$$ which happens only once to cache the time bounds).

We pay an extra $$|4 * T| $$ bytes of memory to store a time array (+ extra cache space) 
This will be much faster if our time slices are relatively small compared to the entire graph.

### Perf Results
Since we don't have exhaustive perf tests it's not clear how much we save end-to-end, but this is the single benchmark I have implemented which has been used in my previous prs (iterating a `DGraph` with 1M events)

**Control (`openDG::main`)**
foo

**Case (`openDG::devsorted_perf`)**
foo

### DIFF
--- a/opendg/graph.py
+++ b/opendg/graph.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import pathlib
 from dataclasses import dataclass
 from functools import cached_property
-from typing import Any, List, Optional, Set, Tuple
+from typing import Any, List, Optional, Set
 
 import pandas as pd
 from torch import Tensor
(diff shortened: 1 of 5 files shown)


opendg/graph.py
```diff
@@ -151,12 +151,16 @@ def edge_feats(self) -> Optional[Tensor]:
         )
 
     @staticmethod
-    def _maybe_max(seq: Tuple[Any, Any]) -> Optional[int]:
-        return max(filter(lambda x: x is not None, seq), default=None)
+    def _maybe_max(a: Any, b: Any) -> Optional[int]:
```
Jacob-Chmura, line 154: This is part of the hot path, so these branch prediction misses were accounting for like 20-30% of the runtime (on this simple workload where all we do is iterate a graph)

PR #94, opened 2025-03-20
Add basic perf utils

### LINKED ISSUE #66: Performance Testing - Sampling Profiler with `flameprof`

Jacob-Chmura:
Close #66 

### Example Usage
fo

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -30,3 +30,4 @@ jobs:
         with:
           args: "format --check"
           changed-files: "true"
+          version-file: "pyproject.toml"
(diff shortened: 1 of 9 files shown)


perf/foo.py
Jacob-Chmura, line 1: This is just a placeholder but we can keep for now so that we remember how the utilities can be used. Will remove down the line

.github/workflows/ruff.yml
Jacob-Chmura, line 1: Should have updated the hooks in a separate pr, but this just updates the remote hook (and our local ruff version) and then ensures we use our local config (in pyproject) when running

Issue #69, opened 2025-03-18
Figure out the output API for a batch


Jacob-Chmura:
What are the pieces of info needed when getting a batch? Currently, we return a `DGraph` view. But, based on the pseudo-code we put together, we really need something like:
```python
for batch in loader:
    src, dst, t, msg = batch
```

Should the batch still be `DGraph` then or should it return these tuples? `DGraph` is more general approach but do we need it?

In case of `NeighorLoader`, we need to be able to get all the nodes in a `DGraph` view. Currently, this doesn't exist... We have to, in the current state, convert to events and iterate through.

PR #96, opened 2025-03-20
Add materialize implementation

### LINKED ISSUE #92: `DGraph` materialize function

Jacob-Chmura:
Close #92 
Address #69

### DIFF
--- a/opendg/_storage/backends/array_backend.py
+++ b/opendg/_storage/backends/array_backend.py
@@ -76,6 +76,29 @@ class DGStorageArrayBackend(DGStorageBase):
                 nodes.update(event_nodes)
         return nodes
 
+    def get_edges(
+        self,
+        start_time: Optional[int] = None,
+        end_time: Optional[int] = None,
+        node_slice: Optional[Set[int]] = None,
+    ) -> Tuple[Tensor, Tensor, Tensor]:
+        if not len(self._events):
+            return (
+                torch.empty(dtype=torch.int64),
+                torch.empty(dtype=torch.int64),
+                torch.empty(dtype=torch.int64),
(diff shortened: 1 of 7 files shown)


opendg/_storage/backends/array_backend.py
```diff
@@ -76,6 +76,29 @@ def get_nodes(
...
+        for i in range(self._lb_time_idx(start_time), self._ub_time_idx(end_time)):
+            event = self._events[i]
+            if isinstance(event, EdgeEvent):
+                if node_slice is None or any(e in node_slice for e in event.edge):
+                    src.append(event.src)
+                    dst.append(event.dst)
+                    t.append(event.t)
+        return torch.Tensor(src), torch.Tensor(dst), torch.Tensor(t)
+
     def get_num_edges(
```
Jacob-Chmura, line 102: In a subsequent PR, I will simplify the backend API by removing these methods (`num_edges`, `num_nodes`) and instead caching the call at the `DGraph` level with the actual `num_edges`

opendg/graph.py
```diff
@@ -163,6 +178,10 @@ def _maybe_min(a: Any, b: Any) -> Optional[int]:
         return a if b is None else b if a is None else None
 
 
+# src, dst, time, features = {'node': node_features, 'edge': dge_features}
```
Jacob-Chmura, line 181: I specifically used a dictionary to pass back the features, because otherwise, it seems easy to mix up the order of node features and edge features. And since computing features (when they don't exist) is a no-op, the user is just free to easily ignore them, instead of having to do things like `src, dst, time, _, edge_features` all of the time

PR #101, opened 2025-03-21
Add node property to `DGraph`

### LINKED ISSUE #99: Add `.nodes` property on `DGraph`

Jacob-Chmura:
Close #99

### DIFF
--- a/test/test_dgraph.py
+++ b/test/test_dgraph.py
@@ -37,8 +37,8 @@ def test_init_empty(time_delta):
     assert dg.node_feats is None
     assert dg.edge_feats is None
 
-    expected = torch.Tensor([]), torch.Tensor([]), torch.Tensor([])
-    _assert_edge_eq(dg.edges, expected)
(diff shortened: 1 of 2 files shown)


test/test_dgraph.py
```diff
@@ -52,13 +53,14 @@ def test_init_from_events(events):
     assert dg.num_nodes == 9
     assert dg.num_edges == 3
     assert dg.num_timestamps == 4
+    assert dg.nodes == {1, 2, 4, 6, 8}
```
Jacob-Chmura, line 56: @shenyangHuang Let's verify these semantics. It's a bit confusing in my opinion. One the one hand, we had advised that `num_nodes` is meant to be the _max_  node id (0-indexed) so that if we have node id 8, we assuming nodes 0 through 7 exist (even if there is no event associated with them).

But then when we actually get the nodes (with events), we can see they don't match the _number of nodes_. What is the right behaviour we want to enforce here.

PR #112, opened 2025-03-23
Add temporal Attention layer

### LINKED ISSUE #111: Temporal Attention Module

Jacob-Chmura:
Close #111

### DIFF
--- a/opendg/nn/__init__.py
+++ b/opendg/nn/__init__.py
@@ -1,4 +1,5 @@
 from .memory import EdgeBankPredictor
 from .time_encoding import Time2Vec
+from .attention import TemporalAttention
 
-__all__ = ['EdgeBankPredictor', 'Time2Vec']
+__all__ = ['EdgeBankPredictor', 'Time2Vec', 'TemporalAttention']
--- a/opendg/nn/attention.py
+++ b/opendg/nn/attention.py
@@ -1,0 +1,76 @@
+import torch
+
+
+class TemporalAttention(torch.nn.Module):
+    r"""Multi-head Temporal Attention Module."""
+
+    def __init__(
+        self,
+        n_heads: int,
+        node_dim: int,
+        edge_dim: int,
+        time_dim: int,
+        out_dim: int,
+        dropout: float = 0.1,
+    ) -> None:
+        super().__init__()
+        if any((x <= 0 for x in [n_heads, node_dim, edge_dim, time_dim, out_dim])):
+            raise ValueError('n_heads,node_dim,edge_dim,time_dim,out_dim must be > 0')
+        if out_dim % n_heads != 0:
+            raise ValueError(f'out_dim: ({out_dim}) % n_heads: ({n_heads}) != 0')
+
+        self.n_heads = n_heads
+        self.head_dim = out_dim // n_heads
+        self.out_dim = out_dim
+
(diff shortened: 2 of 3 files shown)


opendg/nn/attention.py
```diff
@@ -0,0 +1,76 @@
...
+
+        A = torch.einsum('bhld,bhnd->bhln', Q, V)  # (batch, n_heads, 1, num_nbrs)
+        A *= self.head_dim**-0.5
+        del Q, K
+
+        nbr_mask = nbr_mask == 0
+        nbr_mask = nbr_mask.reshape(nbr_mask.shape[0], 1, 1, -1)
+        nbr_mask = nbr_mask.repeat(1, self.n_heads, 1, 1)
+
+        # If a node has no neighbors (nbr_mask all zero), setting masks to -np.inf will cause softmax nans
```
Jacob-Chmura, line 63: https://github.com/yule-BUAA/DyGLib/blob/3aacc36b94b8d2d8293d70a74fdf6d39089b4163/models/modules.py#L184

opendg/nn/__init__.py
```diff
@@ -1,4 +1,5 @@
 from .memory import EdgeBankPredictor
 from .time_encoding import Time2Vec
+from .attention import TemporalAttention
 
-__all__ = ['EdgeBankPredictor', 'Time2Vec']
+__all__ = ['EdgeBankPredictor', 'Time2Vec', 'TemporalAttention']
```
shenyangHuang, line 5: actually does it make sense to have Time2Vec in the same folder as EdgeBank and Temporal attention?
maybe time encoding should have its own folder. We can quickly discuss folder layout tomorrow

opendg/nn/attention.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(
+        self,
+        n_heads: int,
+        node_dim: int,
+        edge_dim: int,
+        time_dim: int,
+        out_dim: int,
+        dropout: float = 0.1,
+    ) -> None:
```
shenyangHuang, line 15: maybe some documentation for input arguments

opendg/nn/attention.py
```diff
@@ -0,0 +1,76 @@
...
+        super().__init__()
+        if any((x <= 0 for x in [n_heads, node_dim, edge_dim, time_dim, out_dim])):
```
shenyangHuang, line 17: might be a good idea to have a `util` function that we reuse to check for integer dimensions for all models. For example, `_check_valid_dims(List[int])`

opendg/nn/attention.py
shenyangHuang: yeah, setting empty node to a very small number is sensible

shenyangHuang (commented):
Looks great overall, left some comments

opendg/nn/__init__.py
Jacob-Chmura: Will merge for now, we can restructure once more layers are in and it's more clear

opendg/nn/attention.py
Jacob-Chmura: Will do

opendg/nn/attention.py
Jacob-Chmura: Completely agree. But I'll add this once we have more layers in, that way we know what we need, and how we can consolidate

PR #113, opened 2025-03-24
Add TGB IO

alip67:
Add the support for the TGB linkproppred to the IO, we could also add tgb/nodeproppred down the line.

### DIFF
--- a/opendg/_io/__init__.py
+++ b/opendg/_io/__init__.py
@@ -1,5 +1,6 @@
 from opendg._io.csv import read_csv
 from opendg._io.pandas import read_pandas
+from opendg._io.tgb import read_tgb
 
 from typing import Any, List
 import pathlib
@@ -14,5 +15,7 @@ def read_events(data: str | pathlib.Path | pd.DataFrame, **kwargs: Any) -> List[
         return read_pandas(data, **kwargs)
     elif isinstance(data, (str, pathlib.Path)):
         return read_csv(data, **kwargs)
+    elif isinstance(data, str) and str(data).startswith('tgb'):
+        return read_tgb(name=str(data), **kwargs)
     else:
         raise ValueError('cannot read events from type {type(data)}')
--- a/opendg/_io/__init__.py
+++ b/opendg/_io/__init__.py
@@ -2,7 +2,7 @@ from opendg._io.csv import read_csv
 from opendg._io.pandas import read_pandas
 from opendg._io.tgb import read_tgb
 
-from typing import Any, List
+from typing import Any, List, Union
 import pathlib
 
 import pandas as pd
@@ -10,12 +10,20 @@ import pandas as pd
 from opendg.events import Event
 
 
-def read_events(data: str | pathlib.Path | pd.DataFrame, **kwargs: Any) -> List[Event]:
+def read_events(
+    data: Union[str, pathlib.Path, pd.DataFrame], **kwargs: Any
+) -> List[Event]:
     if isinstance(data, pd.DataFrame):
         return read_pandas(data, **kwargs)
-    elif isinstance(data, (str, pathlib.Path)):
-        return read_csv(data, **kwargs)
-    elif isinstance(data, str) and str(data).startswith('tgb'):
-        return read_tgb(name=str(data), **kwargs)
-    else:
-        raise ValueError('cannot read events from type {type(data)}')
+
+    if isinstance(data, (str, pathlib.Path)):
+        data_str = str(data)
+
+        if data_str.startswith('tgbl-'):
+            return read_tgb(name=data_str, **kwargs)
+        if data_str.endswith('.csv'):
+            return read_csv(data, **kwargs)
+
+        raise ValueError(f'Unsupported file format or dataset identifier: {data_str}')
+
+    raise ValueError(f'Cannot read events from type {type(data).__name__}')
--- a/opendg/_io/tgb.py
+++ b/opendg/_io/tgb.py
@@ -2,7 +2,6 @@ from typing import Any, List
 
 import torch
 from tgb.linkproppred.dataset import LinkPropPredDataset
-from tgb.nodeproppred.dataset import NodePropPredDataset
 
 from opendg.events import EdgeEvent, Event
 
@@ -16,7 +15,7 @@ def read_tgb(
     if name.startswith('tgbl-'):
         dataset = LinkPropPredDataset(name=name, **kwargs)
     elif name.startswith('tgbn-'):
-        dataset = NodePropPredDataset(name=name, **kwargs)
+        raise ValueError(f'Not Implemented dataset: {name}')
     else:
         raise ValueError(f'Unknown dataset: {name}')
     data = dataset.full_data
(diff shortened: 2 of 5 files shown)


opendg/_io/tgb.py
```diff
@@ -0,0 +1,45 @@
...
+        events.append(event)
+    return events
+
+
+def main():
+    name = "tgbl-wiki"
+    read_tgb(name=name)
+
+if __name__ == "__main__":
+    main()
```
Jacob-Chmura, line 45:
```suggestion
```

We need to hook this up from `opendg/_io/__init__.py` as I had previously mentioned here: https://github.com/shenyangHuang/openDG/blob/98b62445ccf4ad5344354fc09425cd82a0cb6579/opendg/_io/__init__.py#L12-L18

opendg/_io/tgb.py
```diff
@@ -0,0 +1,45 @@
...
+
+from opendg.events import EdgeEvent, Event
+
+
+def read_tgb(
+    name: str,
+) -> List[Event]:
+    # TODO: Node Events not supported
+    events: List[Event] = []
+    dataset = LinkPropPredDataset(name=name, root="datasets", preprocess=True)
```
Jacob-Chmura, line 16: Please pass the `root` and `preprocess` as keword arguments so that they can be forwarded through the `DGraph` constructor: https://github.com/shenyangHuang/openDG/blob/98b62445ccf4ad5344354fc09425cd82a0cb6579/opendg/graph.py#L33

opendg/_io/tgb.py
Jacob-Chmura, line 1: Did you install pre-commit? The formatting and type check CI is failing

opendg/_io/tgb.py
```diff
@@ -0,0 +1,45 @@
...
+    data = dataset.full_data
```
Jacob-Chmura, line 17: Might be worth adding a keyword argument that we pass through, to decide whether we take the train/test or all splits, defaulting to `'all'` as you have done here

test/test_io/test_tgb_io.py
Jacob-Chmura, line 1: The tests are failing in CI please address

test/test_io/test_tgb_io.py
```diff
@@ -0,0 +1,35 @@
...
+        assert events[i].edge == (int(sources[i]), int(destinations[i]))
+
+
+
+def main():
+    
+    test_tgb_conversion()
+
+if __name__ == "__main__":
+    main()
```
Jacob-Chmura, line 35:
```suggestion
```

opendg/_io/tgb.py
alip67: I add the split argument in the defintion, and refactor the whole read_tgb

test/test_io/test_tgb_io.py
alip67: It should be fine right now, I totally missed that part

test/test_io/test_tgb_io.py
alip67: fixed

opendg/_io/tgb.py
alip67: I fixed the **kwargs and also add support for loading the TGBN datasets for the future

opendg/_io/tgb.py
alip67: Fixed, it should be ok right now. I refactored the whole class to make it more clear

opendg/_io/tgb.py
alip67: fixed

opendg/_io/__init__.py
```diff
@@ -14,5 +15,7 @@ def read_events(data: str | pathlib.Path | pd.DataFrame, **kwargs: Any) -> List[
         return read_pandas(data, **kwargs)
     elif isinstance(data, (str, pathlib.Path)):
```
Jacob-Chmura, line 16: As I mentioned in one our previous conversations, this won't work as written. We should put the below branch first, or ensure that the data ends with `'.csv'` before trying to read a csv. Otherwise, there is ambiguity here.

test/test_io/test_tgb_io.py
```diff
@@ -0,0 +1,20 @@
+from tgb.linkproppred.dataset import LinkPropPredDataset
+
+from opendg._io.tgb import read_tgb
+from opendg.events import EdgeEvent
+
+
+def test_tgb_conversion():
+    name = 'tgbl-wiki'
+    dataset = LinkPropPredDataset(name=name, root='datasets', preprocess=True)
```
Jacob-Chmura, line 9: foo

We need to mock this so that the unit test does not depend on downloading a huge dataset everytime we run the tests. If we want actually to test the data download end-to-end, we should put it in integration tests. 

The right way to do this, is to mock the tgb client, and ensure it is being called with the right arguments.

opendg/_io/tgb.py
```diff
@@ -0,0 +1,53 @@
...
+def read_tgb(
+    name: str,
+    split: str = 'all',  # Options: 'train', 'valid', 'test', 'all'
+    **kwargs: Any,
+) -> List[Event]:
+    # TODO: Node Events not supported
+    if name.startswith('tgbl-'):
+        dataset = LinkPropPredDataset(name=name, **kwargs)
+    elif name.startswith('tgbn-'):
+        dataset = NodePropPredDataset(name=name, **kwargs)
```
Jacob-Chmura, line 19: Do we support this right now? I thought not. If it works, then we can keep it

Jacob-Chmura (changes requested):

opendg/_io/tgb.py
shenyangHuang: So as it stands right now, the `NodeEvents` are ignored but the `EdgeEvents` would be loaded. But also because of that the dataset is loaded incompletely. We should create a separate issue and PR to merge for the `NodePropPred` datasets. Maybe throw an error for now. not implemented error

opendg/_io/tgb.py
shenyangHuang: Yeah, I agree, this is a good idea.

test/test_io/test_tgb_io.py
shenyangHuang: I think this should be integration test as well. 
We should test for behaviors such as incorrect dataset name `tgbl-fake` or sth here as well, to see if error is raised correctly

alip67: Fix this error.

test/test_io/test_tgb_io.py
```diff
@@ -1,20 +1,43 @@
...
+
+    # Create mock dataset
+    mock_dataset = MagicMock()
+    mock_dataset.full_data = {
+        'sources': sources,
+        'destinations': destinations,
+        'timestamps': timestamps,
+        'edge_feat': edge_feat,
+    }
+    mock_dataset.train_mask = np.ones(num_events, dtype=bool)
```
Jacob-Chmura, line 27: Maybe try checking that this works when the user asks for `.train`, `.test` splits etc

alip67: Added mock test for different splits.

Issue #149, opened 2025-03-28
Diagnose Bogging Down


Jacob-Chmura:
Could be cache eviction issue, or just more possible historical neighbours when further in time.

Need to check memory profiling to see if memory leak exist

PR #150, opened 2025-03-28
TGAT Throughput

Jacob-Chmura:
#149


opendg/loader/neighbor_loader.py
```diff
@@ -58,6 +61,9 @@ def pre_yield(self, batch: DGraph) -> DGraph:
             for node, _ in seed_nbrhood[-1]:  # Only care about final hop
                 temporal_nbrhood.add(node)  # Don't care about time info either
 
-        batch = self._dg.slice_time(end_time=batch.end_time)
```
Jacob-Chmura, line 61: Problem is that with lower bound on time, over time, the amount we can binary search decreases, since our upper/lower bound time grows (making this effectively linear)

Jacob-Chmura: The question becomes, can we efficiently keep a lower bound on start time after slice neighbours? Currently, we evict our cache, and the next batch basically has to assume no lower bound on start time.

Technically, we could sample a neighbour all the way from origin time. But in practice (especially for biased weighted sampler), most should be very recent.

At the extreme case of the recency sampler, we **need** to make sure we can maintain this information in $\mathcal{0}(1)\$

Jacob-Chmura: Of course, this could be just something inheret to the array backend, and we shouldn't spend too much time fighting it if it's never going to be efficient.

PR #160, opened 2025-03-29
Force materialize and writeable dbatch

### LINKED ISSUE #159: Make `DGBatch` writeable and force materialize

Jacob-Chmura:
Close #159

### DIFF
--- a/opendg/graph.py
+++ b/opendg/graph.py
@@ -194,7 +194,7 @@ class DGraph:
         return a if b is None else b if a is None else None
 
 
-@dataclass(slots=True)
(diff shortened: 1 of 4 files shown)


opendg/graph.py
```diff
@@ -194,7 +194,7 @@ def _maybe_min(a: Any, b: Any) -> Optional[int]:
         return a if b is None else b if a is None else None
 
 
-@dataclass(slots=True)
```
Jacob-Chmura, line 197: Obviously minor perf hit for dropping slots and having `__dict__`

PR #190, opened 2025-04-19
adding time granularity to tgb datasets

shenyangHuang:
instead of adding time coarsen function, I believe currently loading by time granularity is sufficent enough for any discretization research need. Instead, I added added time granularity for `tgbl-wiki` and `tgbl-review` now. 

- [x] adding time granularity to TGB datasets (partially done for now, enough for initial testing
- [x] adding GCN example to run on materialized hourly snapshots

### DIFF
--- a/opendg/graph.py
+++ b/opendg/graph.py
@@ -8,7 +8,7 @@ from typing import Any, List, Optional, Set, Tuple
 import pandas as pd
 from torch import Tensor
 
-from opendg._io import read_events
+from opendg._io import read_events, read_time_delta
 from opendg._storage import DGSliceTracker, DGStorage
 from opendg.events import Event
 from opendg.timedelta import TimeDeltaDG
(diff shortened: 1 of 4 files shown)


Jacob-Chmura (commented):
Nice work

shenyangHuang:
Note that GCN achieves 85%+ AUC and AP after 10epochs.

shenyangHuang:
@Jacob-Chmura  Note that PyTorch Geometric temporal has high dependency on PyG, we might need to have PyG dependency

opendg/graph.py
```diff
@@ -35,6 +35,10 @@ def __init__(
                 raise ValueError('Tried to initialize a DGraph with empty events list')
             self._storage = DGStorage(events)
 
+        # load the correct time granularity if loading TGB
+        if isinstance(data, str) and data.startswith('tgb'):
+            self.time_delta = read_time_delta(data)
```
Jacob-Chmura, line 40: In the current implementation, the `time_delta` that the user passes through the constructor will be overridden by the TGB time delta.

Is there a plan to enable customization here? I'm thinking if the TGB time delta is seconds, and user explicitly specifies a time delta, we should attempt to do the conversion under the hood (if it makes sense)

Jacob-Chmura (approved):

PR #193, opened 2025-05-16
adding GCLSTM example

shenyangHuang:
- [ ] add GCLSTM layer in `nn/recurrent/gclstm.py`
- [ ] add example of running GCLSTM
- [ ] GCLSTM runs (setup similar to GCN) achieves 85% AUC 

to do 
- [ ] clean code for GCLSTM (currently mainly original code)
- [ ] check how to best call layer names etc.

### DIFF
--- a/examples/gclstm_example.py
+++ b/examples/gclstm_example.py
@@ -1,0 +1,232 @@
+"""code adapted from:
+https://github.com/shenyangHuang/UTG/blob/main/dtdg_gclstm.py
+mlp decoder architecture has been changed
+#! can support edge weights check with DGBatch
+"""
+
+import argparse
+from pprint import pprint
+from typing import Tuple
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.nn import Linear
+from torchmetrics import Metric, MetricCollection
+from torchmetrics.classification import BinaryAUROC, BinaryAveragePrecision
+from tqdm import tqdm
+
+from opendg.graph import DGBatch, DGraph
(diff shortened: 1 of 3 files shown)


opendg/nn/recurrent/__init__.py
Jacob-Chmura, line 1: We should also export this as public from the parent module

opendg/nn/recurrent/gclstm.py
```diff
@@ -0,0 +1,243 @@
+"""implmentation adapted from
+https://github.com/benedekrozemberczki/pytorch_geometric_temporal/blob/master/torch_geometric_temporal/nn/recurrent/gc_lstm.py
+"""
+
+from typing import Tuple
+
+import torch
+from torch.nn import Parameter
+from torch_geometric.nn import ChebConv
```
Jacob-Chmura, line 9: This is not currently a dependancy

examples/gclstm_example.py
```diff
@@ -0,0 +1,232 @@
...
+    batch_size=1,
+    batch_unit=args.time_gran,
+)
+
+if train_dg.node_feats_dim is not None:
+    raise ValueError(
+        'node features are not supported yet, make sure to incorporate them in the model'
+    )
+
+#! we need to add static node features to DGraph
```
Jacob-Chmura, line 209: This should be it's own PR, and requires adapting our io as well

Jacob-Chmura (commented):
Thanks for the start, I will continue with this

PR #227, opened 2025-05-29
Add DGData in IO

### LINKED ISSUE #230: Integrate DGData in IO

Jacob-Chmura:
Waiting to merge upstream of #229 and #226 

Close #230
Stack #229


opendg/_io/csv.py
```diff
@@ -13,23 +13,25 @@ def read_csv(
     dst_col: str,
     time_col: str,
     edge_feature_col: Optional[List[str]] = None,
-) -> List[Event]:
+) -> DGData:
     # TODO: Node Events not supported
+
     file_path = str(file_path) if isinstance(file_path, pathlib.Path) else file_path
-    events: List[Event] = []
+    edge_index_, timestamps_, edge_feats_ = [], [], []
```
shenyangHuang, line 20: maybe there is no speed up. But I am wondering if it is faster to find out how many lines are in CSV and just make a large numpy array or torch tensor and then populate it line by line rather than appending and then converting

shenyangHuang (approved):
Looks good, good to merge after the other two PRs are merged

opendg/_io/csv.py
Jacob-Chmura: Good idea, doubt its faster

Jacob-Chmura: So, might as well do it I suppose

PR #238, opened 2025-06-03
Dynamic Node Features

### LINKED ISSUE #237: Add dynamic node events to `DGData`

Jacob-Chmura:
Close #237


opendg/_io/csv.py
```diff
@@ -36,4 +36,6 @@ def read_csv(
             for j, col in enumerate(edge_feature_col):
                 edge_feats[i, j] = float(row[col])  # type: ignore
 
-    return DGData(edge_index, timestamps, edge_feats)
+    return DGData.from_raw(
```
shenyangHuang, line 39: why do we need a separate `from_raw` method?

opendg/data.py
```diff
@@ -1,55 +1,194 @@
...
+        if not isinstance(self.timestamps, Tensor):
+            raise TypeError('timestamps must be a Tensor')
+        if (
+            self.timestamps.ndim != 1
+            or self.timestamps.shape[0] != num_edges + num_node_events
+        ):
+            raise ValueError('timestamps must have shape [num_edges + num_node_events]')
 
         # Sort if necessary
         if not torch.all(torch.diff(self.timestamps) >= 0):
```
shenyangHuang, line 110: actually here we don't check for negative timestamps, maybe we should enforce timestamps >= 0

shenyangHuang (commented):
Looks fine, happy to try `nodeproppred` next

PR #241, opened 2025-06-03
Adding draft for TGB nodeproppred task

shenyangHuang:
- [x] added TGB node label loading logic
- [x] test case with `tgbn-trade`
- [x] add test case with no real data loading needed
- [ ] add TGN running example (won't fix in this PR, leave for future)
- [x] add GCN example running on TGB node prop pred 
- [x] need to update TGB version to support retrieving the `node_label_dict` directly
- [x] add node feat related info into `DGBatch`


opendg/data.py
```diff
@@ -392,8 +393,34 @@ def _check_tgb_import() -> 'LinkPropPredDataset':  # type: ignore
...
+            node_label_dim = 0
+            for t in node_label_dict:
+                for node_id, label in node_label_dict[t].items():
+                    num_node_events += 1
+                    node_label_dim = label.shape[0]
+            node_timestamps = np.zeros(num_node_events, dtype=np.int64)
+            node_ids = np.zeros(num_node_events, dtype=np.int64)
+            dynamic_node_feats = np.zeros(
+                (num_node_events, node_label_dim), dtype=np.float32
+            )
```
Jacob-Chmura, line 413: Nit, but can't we just direcltly create tensors instead of numpy arrays and then converting?

test/test_data.py
```diff
@@ -612,3 +612,55 @@ def test_from_tgb_with_static_node_features(
...
+    mock_dataset.val_mask = val_mask
+    mock_dataset.test_mask = test_mask
+    mock_dataset.num_edges = num_events
+
+    mock_dataset_cls.return_value = mock_dataset
+
+    data = DGData.from_tgb(name='tgbn-trade', split=split)
+    assert isinstance(data, DGData)
+    data.edge_index.tolist()
+    data.timestamps.tolist()
```
Jacob-Chmura, line 665: Are we missing assertions here to ensure ensure they are as expected. And most importantly, that dynamic features are correctly processed.

examples/nodeproppred/gcn.py
```diff
@@ -0,0 +1,236 @@
+import argparse
+import time
+from typing import Tuple
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from tgb.nodeproppred.evaluate import Evaluator
```
Jacob-Chmura, line 8: How does this work? What are the tradeoffs from using tgb here vs. implementing the evaluation ourselves?

Jacob-Chmura (changes requested):
This mostly looks good. Left some comments. There will be some minor merge conflicts with  https://github.com/shenyangHuang/openDG/pull/252

Can we also run the formatter, fix the failing tests, and mypy.

examples/.DS_Store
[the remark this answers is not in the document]
shenyangHuang: done

test/test_data.py
shenyangHuang: adding assertion, now has the magicmock problem

opendg/graph.py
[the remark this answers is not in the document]
shenyangHuang: done

PR #253, opened 2025-06-07
Simplify Unit Tests

### LINKED ISSUE #247: Simplify and refine unit tests

Jacob-Chmura:
Goal here is to keep coverage but simplify test cases. Adding missing test cases will be handled separately.

Close #247

### DIFF
--- a/test/test_storage_impl.py
+++ b/test/test_storage_impl.py
@@ -39,19 +39,7 @@ def edge_only_data_with_features():
 
 
 @pytest.fixture
-def data_with_multi_events_per_timestamp():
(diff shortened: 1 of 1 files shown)


opendg/loader.py
```diff
@@ -34,8 +34,6 @@ def __init__(
     ) -> None:
         if batch_size <= 0:
             raise ValueError(f'batch_size must be > 0 but got {batch_size}')
-        if not len(dg):
```
Jacob-Chmura, line 37: We validate input on construction, so we can assume by now that the graph is valid (non-empty). We should review all user endpoints and ensure all our validation is done there before first release.

